PR #70, opened 2020-03-07
Allow for default tools to be copied from interface folder

valentinptc:
add __dirname and your tool name to the addNode function. This will copy the default tool that is to be used within the tool used for your nodes.


valentinptc:
Just figured out that this should be solved a different way. Namely the web-server should route to the interface/tools/<tool> and not copy and potentially overwrite existing work.

PR #86, opened 2020-03-10
Remove change case

hobinjk-ptc:
Ends up incorporating the eslint changes but I'll rebase after that's approved

### DIFF
--- a/libraries/webFrontend.js
+++ b/libraries/webFrontend.js
@@ -45,7 +45,6 @@
 
 var utilities = require('./utilities');
 var fs = require('fs');
-var changeCase = require('change-case');
 var debug = false;
 var path = require('path');
 var hardwareAPI = require('./hardwareInterfaces');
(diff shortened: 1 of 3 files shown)


libraries/webFrontend.js
```diff
@@ -694,7 +693,7 @@ exports.uploadTargetContent = function (parm, objectsPath, objectInterfaceName)
 
             if (content[1][0] !== '.' && content[1][0] !== '_') {
                 if (debug)console.log(content[1]);
-                var fileTypeF = changeCase.lowerCase(content[1].split('.')[1]);
+                var fileTypeF = content[1].split('.')[1].toLowerCase();
```
benptc: nice. not sure why we needed a module to do that before.

hobinjk-ptc: change-case dropped their lowerCase function so I think they agree

PR #89, opened 2020-03-11
Set tool to interface tools path

valentinptc:
new API for hardware Interfaces
  server.setTool("object", "frame", "toolfolderName", __dirname);
  server.clearTool("object", "frame");

Use this API if you want to deliver a default local tool with your interface.
Save your default tool in interfaces/interfaceName/tools/toolName/

### DIFF
--- a/libraries/hardwareInterfaces.js
+++ b/libraries/hardwareInterfaces.js
@@ -328,6 +328,64 @@ exports.triggerUDPCallbacks = function(msgContent) {
     });
 };
 
+exports.clearTool = function(objectName, frameName){
+    var objectID = utilities.getObjectIdFromTarget(objectName, objectsPath);
+    console.log('remove set tool');
+
+    var frameUuid = objectID + frameName;
+
+    if (!_.isUndefined(objectID) && !_.isNull(objectID)) {
+        if (objects.hasOwnProperty(objectID)) {
+            if (objects[objectID].frames.hasOwnProperty(frameUuid)) {
+                if (objects[objectID].frames[frameUuid].hasOwnProperty('tool')) {
+                    delete objects[objectID].frames[frameUuid].tool;
+                }
+            }
+        }
+    }
+}
+
+exports.setTool = function(objectName, frameName, tool, dirName){
+  
+    var objectID = utilities.getObjectIdFromTarget(objectName, objectsPath);
+    console.log('set new tool: ', tool);
+
+    var frameUuid = objectID + frameName;
+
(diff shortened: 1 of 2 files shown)


libraries/hardwareInterfaces.js
```diff
@@ -328,6 +328,64 @@ exports.triggerUDPCallbacks = function(msgContent) {
...
 };
 
+exports.clearTool = function(objectName, frameName){
+    var objectID = utilities.getObjectIdFromTarget(objectName, objectsPath);
+    console.log('remove set tool');
+
+    var frameUuid = objectID + frameName;
+
+    if (!_.isUndefined(objectID) && !_.isNull(objectID)) {
+        if (objects.hasOwnProperty(objectID)) {
```
hobinjk-ptc: These two lines could be simplified as just `if (objects.hasOwnProperty(objectID)) {` since `objects` will never contain a null or undefined objectID

server.js
```diff
@@ -1111,44 +1112,79 @@ function objectWebServer() {
...
         urlArray.splice(0, 1);
         urlArray.splice(0, 1);
         if (urlArray[1] === 'frames') {
+            var objectKey = utilities.readObject(objectLookup, urlArray[0]);
+            var frameKey = utilities.readObject(objectLookup, urlArray[0]) + urlArray[2];
+            var thisFrame = getFrame(objectKey, frameKey);
+            
+            var toolpath = null;
+            
+            if(thisFrame !== null){
```
hobinjk-ptc: the !== null is redundant

server.js
```diff
@@ -1111,44 +1112,79 @@ function objectWebServer() {
...
 
         var newUrl = '';
+        var newToolUrl = '';
         for (let i = 0; i < urlArray.length; i++) {
             newUrl += '/' + urlArray[i];
+            
         }
 
+        if(toolpath !== null) {
+            for (let i = 2; i < urlArray.length; i++) {
```
hobinjk-ptc: It would be nice to get comments explaining why numbers like this 2 and the above -1 -2 -3 come from since if there is a future attempt to touch this code or refactor it to be more streamlined every single magic number will require thought and reverse-engineering.

server.js
```diff
@@ -1111,44 +1112,79 @@ function objectWebServer() {
...
+                newToolUrl += '/' + urlArray[i];
+            }
+        }
+        
         if (newUrl.slice(-1) === '/') {
             newUrl += 'index.html';
+            if(toolpath !== null) {
```
hobinjk-ptc: Another opportunity to simplify to `if (toolpath)`

PR #90, opened 2020-03-12
Auto-initialize world objects without targets when created on frontend

benptc:
Instead of making them deactivated until you add a DAT/JPG target. Slight updates to UI to make it clearer that world objects are enabled by default but you can add an origin target to them:

Screen Shot 2020-03-11 at 4 36 32 PM

### DIFF
--- a/libraries/webInterface/gui/index.js
+++ b/libraries/webInterface/gui/index.js
@@ -1472,16 +1472,16 @@ realityServer.gotClick = function (event) {
                                         });
                                     }, 100);
                                 }
-                            }, 'name='+msgContent.name+'&width='+defaultSize+'&height='+defaultSize);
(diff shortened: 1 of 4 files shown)


libraries/webInterface/gui/index.js
```diff
@@ -1424,17 +1440,48 @@ realityServer.gotClick = function (event) {
...
+                                    setTimeout(function() {
+                                        realityServer.sendRequest('/object/' + msgContent.id + '/activate/', 'GET', function (state) {
+                                            if (state === 'ok') {
+                                                realityServer.objects[msgContent.id].active = true;
+                                            }
+                                            realityServer.update();
+                                        });
+                                    }, 100);
+                                }
+                            }, 'name='+msgContent.name+'&width='+defaultSize+'&height='+defaultSize);
```
hobinjk-ptc: It would be nice to have this spaced out for readability

PR #95, opened 2020-03-17
adds /subscribe/interfaceSettings socket to config.html pages and..

benptc:
.. pushSettingsToGui API to hardware interfaces

Allows the "Manage Hardware Interface" config.html page to subscribe to live updates of the settings, if the hardware interface updates any of the settings after it loads.

How to use:
In your hardware interface, if you update the exports.settings, call:
```
server.pushSettingsToGui('name of your hardware interface', exports.settings);
```

In your config.html, include config.js:
```
<script src="../libraries/gui/config.js"></script>
```
Then create an InterfaceConfig, and subscribe to settings:
```
let interfaceConfig = new InterfaceConfig(realityServer.hardwareInterfaceName);
interfaceConfig.addSettingsUpdateListener(function(thisMsg) {
    let settingName = 'randomNumber';
    let settingValue = thisMsg.currentSettings[settingName].value;
    console.log('new value for ' + settingName + ' is ' + settingValue);
});
```

Note: calling pushSettingsToGui doesn't actually write the exports.settings to the settings.json file on disk. The only way to write to that is from the web frontend. But this can be used to push temporary settings states to the frontend, such as whether the hardware connected successfully.

### DIFF
--- a/libraries/webInterface/gui/config.js
+++ b/libraries/webInterface/gui/config.js
@@ -1,3 +1,11 @@
+/**
+ * @fileOverview
+ * This is a library that should be included in hardware interface's config.html pages in order
+ * to automatically add a websocket to the page and give them access to teh InterfaceConfig
+ * APIs, which will allow them to subscribe to realtime updates to the settings from the
+ * hardware interface's index.js
+ */
+
 let socketIoScript = {};
 let socketIoRequest = {};
 
@@ -21,8 +29,8 @@ function loadScriptSync(url, requestObject, scriptObject) {
 loadScriptSync('../../socket.io/socket.io.js', socketIoRequest, socketIoScript);
 
 /**
- * This is used to allow the config.html pages for hardware interfaces to subscribe to new values
- * @param interfaceName
+ * This is used to allow the config.html pages for hardware interfaces to subscribe to new settings
+ * @param {string} interfaceName - exact name of the hardware interface
  * @constructor
  */
 function InterfaceConfig(interfaceName) { // eslint-disable-line no-unused-vars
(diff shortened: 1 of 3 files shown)


libraries/webInterface/gui/config.js
```diff
@@ -0,0 +1,59 @@
...
+ * hardware interface's index.js
+ */
+
+let socketIoScript = {};
+let socketIoRequest = {};
+
+// Load socket.io.js synchronous so that it is available by the time the rest of the code is executed.
+function loadScriptSync(url, requestObject, scriptObject) {
+    requestObject = new XMLHttpRequest();
+    requestObject.open('GET', url, false);
```
hobinjk-ptc: This may simplify the control flow of the page but it will destroy performance and potentially introduce bugs in Firefox and other browsers which have deprecated synchronous XHRs

It would be much safer to force the config.html pages to also include socket.io since they're already including this file or set up a less brittle asynchronous approach.

benptc: Ah ok. I just copy-and-pasted from how the sockets for the logic block settings pages were set up. I should probably change those too to the async method.

hobinjk-ptc: Oh, hmm. Yeah, that would be a nice project for the future since doing this sort of async thing adds a full roundtrip to the page load time. It's definitely not critical right now since everything is on local networks and very powerful servers

benptc: I updated it to create pending stubs for the socket APIs and trigger them when the script asynchronously loads, as we do in object.js. When I get the chance I can make similar changes to the logic node settings

benptc:
> Nice! This LGTM

Thanks for the review! 🙏

PR #98, opened 2020-03-23
Changed all API calls that contain the word frame to tool

valentinptc:
I adjusted all APIs to rename frame to tool and reality to spatial.
@benptc can you test these APIs with all tools and the hardware interfaces?
Everything should be backward compatible.

### DIFF
--- a/libraries/hardwareInterfaces.js
+++ b/libraries/hardwareInterfaces.js
@@ -83,18 +83,18 @@ function EmptyNode(nodeName, type) {
  * @param {value} value The value to be passed on
  * @param {string} mode specifies the datatype of value, you can define it to be whatever you want. For example 'f' could mean value is a floating point variable.
  **/
-exports.write = function (objectName, frameName, nodeName, value, mode, unit, unitMin, unitMax) {
+exports.write = function (object, tool, node, value, mode, unit, unitMin, unitMax) {
 
-    if (typeof mode === 'undefined')  mode = 'f';
-    if (typeof unit === 'undefined')  unit = false;
-    if (typeof unitMin === 'undefined')  unitMin = 0;
-    if (typeof unitMax === 'undefined')  unitMax = 1;
+    if (typeof mode === 'undefined') mode = 'f';
+    if (typeof unit === 'undefined') unit = false;
+    if (typeof unitMin === 'undefined') unitMin = 0;
+    if (typeof unitMax === 'undefined') unitMax = 1;
 
-    var objectKey = utilities.readObject(objectLookup, objectName); //get globally unique object id
+    var objectKey = utilities.readObject(objectLookup, object); //get globally unique object id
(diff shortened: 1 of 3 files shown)


benptc:
OK, sounds good. I'll test these when I get a chance, before I approve it.

libraries/hardwareInterfaces.js
```diff
@@ -440,44 +459,44 @@ exports.addNode = function (objectName, frameName, nodeName, type, position) {
...
 
-            if (!hardwareObjects.hasOwnProperty(objectName)) {
-                hardwareObjects[objectName] = new EmptyObject(objectName);
+            if (!hardwareObjects.hasOwnProperty(object)) {
+                hardwareObjects[object] = new EmptyObject(object);
             }
 
-            if (!hardwareObjects[objectName].frames.hasOwnProperty(frameUuid)) {
-                hardwareObjects[objectName].frames[frameUuid] = new EmptyFrame(frameName);
+            if (!hardwareObjects[object].rames.hasOwnProperty(frameUuid)) {
```
hobinjk-ptc, line 479: This is the first bug encountered when trying to run the server

libraries/hardwareInterfaces.js
```diff
@@ -381,11 +400,11 @@ exports.setTool = function(objectName, frameName, tool, dirName){
...
-                objects[objectID].frames[frameUuid].tool = {addon: addonName, interface: interfaceName, tool: tool};
+                objects[objectID].frames[frameUuid].tool = {addon: addonName, interface: interfaceName, tool: newTool};
             }
         }
     }
-}
+};
 
 /**
  * @desc addIO() a new IO point to the specified RealityInterface
```
hobinjk-ptc, line 410: this jsdoc comment is now out of date

libraries/hardwareInterfaces.js
```diff
@@ -344,14 +363,14 @@ exports.clearTool = function(objectName, frameName){
...
         }
     }
-}
+};
 
-exports.setTool = function(objectName, frameName, tool, dirName){
-  
-    var objectID = utilities.getObjectIdFromTarget(objectName, objectsPath);
-    console.log('set new tool: ', tool);
+exports.setTool = function (object, tool, newTool, dirName) {
```
hobinjk-ptc, line 368: the naming of tool and newTool here suggests that they're of the same data type/format but the code below appears to use them in different ways

hobinjk-ptc: this doesn't matter in the grand scheme of things

libraries/hardwareInterfaces.js
hobinjk-ptc: (fixed in the commit I just pushed)

hobinjk-ptc:
Merging this because it's going to break my other PR

PR #103, opened 2020-04-03
restart server programatically

vheun:
I found a nice way to restart the server programmatically.
@benptc I think the next step would be to add the functionality you had with the HMI for when the server restarts, the web frontend restarts as well.

### DIFF
--- a/libraries/utilities.js
+++ b/libraries/utilities.js
@@ -61,13 +61,23 @@ var ip = require('ip');       // get the device IP address library
 var dgram = require('dgram'); // UDP Broadcasting library
 var os = require('os');
 var path = require('path');
-
+const { spawn } = require('child_process');
 
 var hardwareInterfaces = {};
 
 var identityFolderName = '.identity'; // TODO: get this from server.js
 var homedir =  path.join(path.join(os.homedir(), 'Documents'), 'realityobjects');
 var hardwareIdentity = homedir + '/.identity';
+
+exports.restartServer = function () {
+  spawn("node", ["server"], {
+    detached: true,
+    stdio: 'inherit'  // [process.stdin, process.stdout, process.stderr]
+  }).unref()
+
+  process.kill(process.pid);
+  process.exit();
+};
 
 exports.writeObject = function (objectLookup, folder, id) {
     objectLookup[folder] = {id: id};
--- a/server.js
+++ b/server.js
@@ -753,6 +753,18 @@ function exit() {
 
 process.on('SIGINT', exit);
 
+
+const readline = require('readline');
+readline.emitKeypressEvents(process.stdin);
+process.stdin.setRawMode(true);
+process.stdin.on('keypress', (str, key) => {
+    if (key.ctrl && key.name == 'c') {
+        hardwareAPI.shutdown();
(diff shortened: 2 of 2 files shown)


server.js
```diff
@@ -753,6 +753,18 @@ function exit() {
 
 process.on('SIGINT', exit);
```
hobinjk-ptc, line 754: This is an existing handler for ctrl-c, it's probably better to see what this function does and modify it rather than create subtly different keypress handler.

libraries/utilities.js
```diff
@@ -61,14 +61,24 @@ var ip = require('ip');       // get the device IP address library
...
 var hardwareIdentity = homedir + '/.identity';
 
+exports.restartServer = function () {
+  spawn("node", ["server"], {
+    detached: true,
+    stdio: 'inherit'  // [process.stdin, process.stdout, process.stderr]
+  }).unref()
+
+  process.kill(process.pid);
+  process.exit();
```
hobinjk-ptc, line 79: What is the difference between these two lines?

server.js
vheun: I would have used this if it would work. Tried every option. Try it out. I have implemented the restart when you click on your network adapter in the web interface.
Once you have done this, this exit method stops working well.
The method I implemented is the one reliable.

libraries/utilities.js
vheun: exit leaves the process, kill kills the app by its pid. kill seems to have a a wider reach when you spawn child processes.

You can remove process.exit(), but then again it never gets called and I left it in.

server.js
vheun: I think this is what happens:
ctrl + C is still associated with the parent process, however that is eliminated and the link to the child is removed. When you call it, you get back to the command line, but the new running process is not eliminated. It still writes to the terminal and its process is still active. If you want to restart the server, you get error messages. 

However, the command line is inherited from the original process and as such you can listen to its input. Since there are no other command line inputs for the server, it seems a feasible solution.

hobinjk-ptc: Yeah, this approach gets the job done at the end of the day.

One behavior that should be noted is that if someone closes their terminal instead of pressing ctrl-c it won't be able to terminate the detached process

In general this approach is hacky enough that it would be nice to get comments on each of the functions describing what they're doing

vheun: Ouch.. I was not testing that. Thats a side effect of not catching SIGINT I guess.
I think this needs more work. We have to find a way how the normal SIGINT behavior works.

vheun: Maybe we can create a start file that is called index.js and server is really just a child process of index.js. index.js is never killed and as such can handle a rightful exit.

benptc: Once upon a time I did create an index.js file that runs server.js as a child process. You can try it out and see if it accomplishes what you're looking for

hobinjk-ptc: See #105, I took the idea of the forever-based index.js and added support for a 'restart' message

valentinptc: perfect!

valentinptc:
This was resolved https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/105

PR #105, opened 2020-04-03
Safe restarts

hobinjk-ptc:
The first commit is the interesting one here. Code in server.js can now call `process.send('restart')` to tell index.js to restart the server. Otherwise everything should behave exactly the same as before


valentinptc:
Everything seems to work.

PR #106, opened 2020-04-03
Further specify that node 6 is only supported for mobile devices

hobinjk-ptc:
Tests are currently failing because express-handlebars snuck in a change incompatible with node 6 but that doesn't impact us since we don't use express-handlebars on mobile


hobinjk-ptc:
Merging, r=me because the failing tests are annoying

PR #107, opened 2020-04-03
Expose setHardwareInterfaceSettings as part of hardware interface API

hobinjk-ptc:


libraries/hardwareInterfaces.js
```diff
@@ -634,10 +634,33 @@ exports.getDebug = function () {
...
  ********** END API FUNCTIONS *********
  */
 
+/**
+ * Complement to setup() which is necessary due to the unique positioning of
+ * the setHardwareInterfaceSettings function
+ */
+exports.setHardwareInterfaceSettingsImpl = function(setHardwareInterfaceSettings) {
+    setHardwareInterfaceSettingsImpl = setHardwareInterfaceSettings;
+};
```
benptc, line 662: Interesting structure but it makes sense why this is needed

hobinjk-ptc: Yeah it's adding a couple of strands of spaghetti

PR #108, opened 2020-04-03
Replace sharp with Jimp

hobinjk-ptc:
This shouldn't be merged yet because there's no reason to jeopardize the upcoming launch with the potential instability of an untested library


valentinptc:
Now we have some more time to test it ;-)

Issue #110, opened 2020-04-03
Toolbox Crashes when network interface changes


valentinptc:
1. Load the Toolbox the first time with wifi off
2. Close the Toolbox
3. change the Wifi to on.
4. Play around with Airplane mode and Wifi on / off

App crashes when we add new tools to the world object.

Why?
The server saves the default network interface at start.
When the network setting change (for example the network interface is removed) the server has an undefined state for its network interface. Hence the Server crashes the app.

How to solve?
Wifi should be the primary network interface to use. We need to check if the network interface is undefined and switch to the one that is available. Once Wifi becomes available again, we need to switch back. At any given moment. IP for the server can not be undefined.

valentinptc:
working on it right now

PR #112, opened 2020-04-04
Fix for critical Bug

valentinptc:
This is a bug-fix for issue #110 

1. Handles network interface changes while the server is running.
2. Uses the next available interface if the defined interface is missing.
3. Prevents a crash if UDP can not send signals
4. Cleaned up all IP related code into one module

### DIFF
--- a/server.js
+++ b/server.js
@@ -85,12 +85,14 @@ const globalVariables = {
 
  */
 
-const serverPort = 8080;
+var serverPort = isMobile ? 49369 : 8080 ;
+const serverUserInterfaceAppPort = 49368;
 const socketPort = serverPort;     // server and socket port are always identical
 const beatPort = 52316;            // this is the port for UDP broadcasting so that the objects find each other.
 const timeToLive = 2;                     // the amount of routers a UDP broadcast can jump. For a local network 2 is enough.
 const beatInterval = 5000;         // how often is the heartbeat sent
 const socketUpdateInterval = 2000; // how often the system checks if the socket connections are still up and running.
+
 
 // todo why would you alter the version of the server for mobile. There should only be one version of the server. 
 // The version of this server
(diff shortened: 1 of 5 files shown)


server.js
```diff
@@ -4483,6 +4488,7 @@ function createObjectFromTarget(objects, folderVar, __dirname, objectLookup, har
             if (objectIDXML.length > 13) {
 
                 objects[objectIDXML] = new ObjectModel(services.ip, serverPort, version, protocol);
```
hobinjk-ptc, line 4490: This still has the serverPort argument

server.js
```diff
@@ -144,12 +147,77 @@ storage.initSync();
...
+var services = {};
+if (!isMobile) {
+    services.networkInterface = require('network-interfaces');
+}
+services.options = {ipVersion: 4};
+services.ips = {activeInterface: null, tempActiveInterface: null, interfaces: {}};
+services.ip = null
+services.updateAllObjcts = function (ip){
+    console.log("updating all objects with new IP: ", ip);
+    for (let key in objects){
```
hobinjk-ptc, line 160: How does this work with objects on other servers?

server.js
```diff
@@ -144,12 +147,77 @@ storage.initSync();
...
+    this.ips.interfaces = {};
+    // if this is mobile, only allow local interfaces
+    if (isMobile) {
+        this.ips.interfaces["mobile"] = '127.0.0.1';
+        this.ips.activeInterface = "mobile";
+        return '127.0.0.1';
+    }
+    
+    // Get All available interfaces
+        var interfaceNames = this.networkInterface.getInterfaces(this.options);
```
hobinjk-ptc, line 174: options can be a local variable instead of stored on this object

server.js
valentinptc: In case the IP for the server changes, you need to make sure that all objects on the server have the right IP. Because this IP is used to send heartbeats and it's sent out to the editors for other references. You just need to make sure that the Object always has the right IP saved.

server.js
valentinptc: sure

server.js
hobinjk-ptc: Oh right, I got confused between this and the userinterface

server.js
valentinptc: Thats a bug

PR #128, opened 2020-04-13
 Fix use of paths with unzipper on Windows

hobinjk-ptc:
1bf2491 is the actual bugfix, the other commit didn't seem necessary but is more consistent.


valentinptc:
this does the trick

PR #132, opened 2020-04-14
Frontend notifications

benptc:
Adds a notification class that can be easily triggered to add an error or message to the top of the screen for a set amount of time or until the X is pressed. testing it out by adding one error notification if you upload a zip file without any target data to the drop zone. Will add more error notifications for other (currently silent) issues in the future.

Screen Shot 2020-04-14 at 11 18 26 AM

### DIFF
--- a/libraries/webInterface/css/notifications.css
+++ b/libraries/webInterface/css/notifications.css
@@ -1,0 +1,34 @@
+.notification {
+    position: relative;
+    left: 0;
+    top: 0;
+    width: 100vw;
+    height: 30px;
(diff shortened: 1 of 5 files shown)


benptc:
Messages stack, fixed to top of screen, regardless of scroll. Errors are red, messages are grey. Here is an exaggerated example of how multiple at the same time look (not beautiful but in reality probably won't have more than 1 or 2 at the same time)

Screen Shot 2020-04-14 at 11 22 43 AM

valentinptc:
nice!

Issue #123, opened 2020-04-09
Add feedback when target zip upload fails


hobinjk-ptc:
https://forum.spatialtoolbox.vuforia.com/t/when-i-upload-the-zip-for-an-object-target-nothing-happens/46

realityLabDiscourseUser:
This issue has been mentioned on **Vuforia Spatial Toolbox Community**. There might be relevant details there:

https://forum.spatialtoolbox.vuforia.com/t/when-i-upload-the-zip-for-an-object-target-nothing-happens/46/6

benptc:
I'm halfway done with a new feature that adds a notification feature where errors and user-facing messages can pop up at the top of the web interface for a few seconds

realityLabDiscourseUser:
This issue has been mentioned on **Vuforia Spatial Toolbox Community**. There might be relevant details there:

https://forum.spatialtoolbox.vuforia.com/t/when-i-upload-the-zip-for-an-object-target-nothing-happens/46/7

PR #133, opened 2020-04-14
adds error message to frontend if try to upload anything other than x…

benptc:
…ml, dat, jpg, or zip to the dropzone. also allows xml upload via dropzone independently from dat, previously you had to upload together or it would be generated automatically

Fixes issue #123 (along with previous PR)

### DIFF
--- a/libraries/webInterface/gui/index.js
+++ b/libraries/webInterface/gui/index.js
@@ -1121,7 +1121,7 @@ realityServer.gotClick = function (event) {
 
             realityServer.myTargetDropzone.on('error', function(file, message) {
                 if (typeof message.error !== 'undefined') {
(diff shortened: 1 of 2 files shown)


benptc:
Message looks like:

Screen Shot 2020-04-14 at 2 04 28 PM

benptc:
> Nice! It would also be good to reject model targets

Do you know which file format model targets are? They would be rejected by this unless they are also packaged as a dat or a zip, in which case I'll need to look into some other distinguishing factor

hobinjk-ptc:
They're also a zip but they contain XML with QCARConfig > Tracking > ModelTarget

benptc:
Created issue #135 to follow up about the model targets

Issue #20, opened 2020-02-10
Sanitize object names before creating them on webFrontend


benptc:
For example, prevent restricted name `_WORLD_local`. Perhaps restrict to alphanumeric characters (at least make sure there are no tabs, newlines, etc getting accidentally added).

PR #134, opened 2020-04-14
adds error message and prevents user from creating objects and frames…

benptc:
… with non-alphanumeric names


benptc:
Shows an error and prevents creation when Add button pressed e.g:

Screen Shot 2020-04-14 at 2 41 09 PM

libraries/webInterface/gui/index.js
```diff
@@ -1571,6 +1568,27 @@ realityServer.gotClick = function (event) {
...
+function isNameOk(textContent) {
+    if (textContent === 'Enter Name') {
+        return false;
+    }
+    if (textContent === '') {
+        return false;
+    }
+    let isAlphanumeric = /^[a-z0-9]+$/i.test(textContent);
+    if (!isAlphanumeric) {
+        showErrorNotification('Name must be alphanumeric');
```
hobinjk-ptc, line 1587: It must be lowercase as well according to the regex

libraries/webInterface/gui/index.js
```diff
@@ -1429,88 +1429,85 @@ realityServer.gotClick = function (event) {
...
-                if (shouldAddWorldObject) {
-                    objectName = '_WORLD_' + textContent;
-                }
+        let objectName = textContent;
+        if (shouldAddWorldObject) {
+            objectName = '_WORLD_' + textContent;
+        }
 
-                // TODO: sanitize object names before creating, for example prevent _WORLD_local (github issue #20)
+        // TODO: sanitize object names before creating, for example prevent _WORLD_local (github issue #20)
```
hobinjk-ptc, line 1445: This is handled by isNameOk as far as I can see

libraries/webInterface/gui/index.js
benptc: oops I'll fix the regex to not be lowercase, I think mixed case should be allowed

libraries/webInterface/gui/index.js
benptc: ah yes, being able to not create non-world objects with the same name as world objects is a non-intended consequence of this

benptc:
Fixes issue #20

PR #136, opened 2020-04-14
adds green success messages to top of screen when upload to dropzone …

benptc:
…works correctly, and fixes upload progress bar reset when needed

target-upload-success-messages 2020-04-14 17_57_06

### DIFF
--- a/.eslintrc-web.js
+++ b/.eslintrc-web.js
@@ -59,6 +59,7 @@ module.exports = {
         'secondMouseDown': 'writable',
         'showErrorNotification': 'writable',
         'showMessageNotification': 'writable',
+        'showSuccessNotification': 'writable',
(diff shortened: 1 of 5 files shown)


realityLabDiscourseUser:

https://forum.spatialtoolbox.vuforia.com/t/when-i-upload-the-zip-for-an-object-target-nothing-happens/46/9

benptc:
> looks good
> 
> Does this responds to the 200 "done" message from the server once the files are uploaded?

Yes, the server responds when the files are uploaded. I didn't actually have to change that part, it was already sending the response, the problem was that the progress bar on the frontend was only resetting in the myTargetDropzone.on('success', ...) callback, not in the myTargetDropzone.on('error', ...) callback, so if the zip failed to upload correctly the progress bar would get stuck

Issue #139, opened 2020-04-17
If an image is uploaded as target, the server gui still asks for a dat file


valentinptc:
I believe this is redundant as we use instant image targets?

realityLabDiscourseUser:
This issue has been mentioned on **Vuforia Spatial Toolbox Community**. There might be relevant details there:

https://forum.spatialtoolbox.vuforia.com/t/missing-dat-file/72/2

PR #140, opened 2020-04-17
better user feedback for when instant image is used

valentinptc:
A better wording and neutral color for when the dat file is not needed. #139


realityLabDiscourseUser:

https://forum.spatialtoolbox.vuforia.com/t/missing-dat-file/72/3

PR #141, opened 2020-04-17
initial server side anchors implementation - we need to test this for a while before we merge

valentinptc:
**The idea:** If World Object has a target and is active, all objects that do not own a target become automatically anchors relative to the world object. Therefore these objects become visible but their position is calculated relative to the world object without a target.

**What this commit does:**
1. The server generates always a full object entry with json file
2. Object have a new key called: isAnchor.
3. If anchor is true, the hard-beat is sent out with a tcs = 0 because there is no file for the checksum.
4. The Frontend uses page reloading to adopt the new anchor rules. Thats a bit lazy on my side, but does the job for now.
4. For the visual frontend, I have defined the World Object as strictly dependent on target information. Anything else will create randomly appearing tools every app start. However we need to make sure that the mobile server still activates its temporary world object in the right way.

With this heartbeat and the new anchor objects, the Spatial Toolbox has to take over from here.


realityLabDiscourseUser:

https://forum.spatialtoolbox.vuforia.com/t/area-targets-and-hue-interface/64/8

libraries/utilities.js
```diff
@@ -270,11 +271,18 @@ var getObjectIdFromTarget = function (folderName, objectsPath) {
                 });
 
         return resultXML;
+    } else if (fs.existsSync(jsonFile)) {
+      let thisObject = JSON.parse(fs.readFileSync(jsonFile, 'utf8'));
```
hobinjk-ptc, line 275: nit: the indentation level here is wrong

libraries/webFrontend.js
```diff
@@ -143,8 +151,8 @@ exports.printFolder = function (objects, objectsPath, debug, objectInterfaceName
         var datExists = fs.existsSync(path.join(objectsPath, objectKey, identityFolderName, '/target/target.dat'));
         var xmlExists = fs.existsSync(path.join(objectsPath, objectKey, identityFolderName, '/target/target.xml'));
         var jpgExists = fs.existsSync(path.join(objectsPath, objectKey, identityFolderName, '/target/target.jpg'));
-
-        if (xmlExists && (datExists || jpgExists)) {
+        
+        if ((xmlExists && datExists && jpgExists) || (xmlExists && jpgExists)) {
```
hobinjk-ptc, line 155: This is equivalent to xmlExists && jpgExists since if the first is true the second will also be true

libraries/webFrontend.js
```diff
@@ -154,10 +162,18 @@ exports.printFolder = function (objects, objectsPath, debug, objectInterfaceName
         }
 
         // world objects are always initialized true regardless of target data
-        if (newObject[thisObjectKey].isWorldObject) {
-            newObject[thisObjectKey].initialized = true;
-        }
 
+            if (newObject[thisObjectKey].isWorldObject) {
```
hobinjk-ptc, line 166: Empty if statement

libraries/webInterface/gui/index.js
```diff
@@ -85,6 +85,43 @@ realityServer.initialize = function () {
     this.initializeHelp();
 };
 
+
+realityServer.activeUiElement = function() {
```
hobinjk-ptc, line 89: Most web APIs would capitalize this activeUIElement

libraries/webInterface/gui/index.js
```diff
@@ -257,6 +294,7 @@ realityServer.updateManageObjects = function(thisItem2) {
                             realityServer.sendRequest('/', 'POST', function(state) {
                                 if (state === 'ok') {
                                     delete realityServer.objects[thisObjectKey];
+                                    window.location.reload();
```
hobinjk-ptc, line 297: You can delete the following realityServer.update

libraries/webInterface/gui/index.js
```diff
@@ -323,7 +361,7 @@ realityServer.updateManageObjects = function(thisItem2) {
                         // only add the icon if it exists
                         if (thisObject.targetsExist.jpgExists) {
                             let ipAddress = realityServer.states.ipAdress.interfaces[realityServer.states.ipAdress.activeInterface];
-                            thisObject.dom.querySelector('.objectTargetIcon').src = 'http://' + ipAddress + ':' + realityServer.states.serverPort + '/obj/' + thisObject.name + '/target/target.jpg';
+                             thisObject.dom.querySelector('.objectTargetIcon').src = 'http://' + ipAddress + ':' + realityServer.states.serverPort + '/obj/' + thisObject.name + '/target/target.jpg';
```
hobinjk-ptc, line 364: wrong indentation

libraries/webFrontend.js
valentinptc: Any of these files can exist or not exist. Two states are allowed:
All files exist or xml and jpg exist.
There is no state where xml and dat is allowed.
The previous expression produced fault results.

libraries/webFrontend.js
valentinptc: why is it empty?

libraries/webFrontend.js
hobinjk-ptc: There's just no reason to have the first half of this expression since if (xmlExists && datExists && jpgExists) is true then (xmlExists && jpgExists) is always also true

libraries/webFrontend.js
hobinjk-ptc: There is no code contained in this if statement so it may as well be deleted or commented out

libraries/webFrontend.js
valentinptc: Sure so we could write it like: 
if (xmlExists && ((datExists && jpgExists)  || jpgExists))

libraries/webInterface/gui/index.js
valentinptc: function does not exist anymore

libraries/webFrontend.js
valentinptc: commented

libraries/webFrontend.js
hobinjk-ptc: No it just simplifies to `xmlExists && jpgExists`

When all files exist the xml and jpg files exist

valentinptc:
I will merge this into the master now. @hobinjk-ptc can you update the electron app from the new master so that we can test this state.

PR #165, opened 2020-05-21
Adding README for Raspberry Pi

SainaRez:


RaspberryPi_README.md
```diff
@@ -0,0 +1,84 @@
...
+
+#### The Following Steps are the same for Ubuntu and Raspian:
+
+When the OS is installed, make sure you install git: 
+
+`$ sudo apt install git`
+
+You can install Node.js v14.x by running the following commands:
+
+`$ curl -sL https://deb.nodesource.com/setup_13.x | sudo -E bash - `
```
hobinjk-ptc, line 42: This should be formatted as the following (dropping the '\' )
```
\```
$ curl -sL https://deb.nodesource.com/setup_13.x | sudo -E bash - 
$ sudo apt-get install -y nodejs 
\```
```
Right now all the code blocks are on one line

SainaRez: Oh is that for all the code snippets?

RaspberryPi_README.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+### Ubuntu:
+
+You can install Ubuntu on the Raspberry Pi using a MicroSD Card.
+First you need to download the OS image that will be written to the SD Card. 
+
+Different versions of Ubuntu 20.04 LTS and 18.04 can be found [here](https://ubuntu.com/download/raspberry-pi).
+
+In order to write the image of Ubuntu, you need to write the OS image onto the MicroSD Card. We used the Win32 Disk Imager softwre for windows.
```
hobinjk-ptc, line 14: typo: 'softwre'

RaspberryPi_README.md
hobinjk-ptc: Yep, the multiline code block with triple backticks is the correct thing to use in those places

RaspberryPi_README.md
```diff
@@ -0,0 +1,99 @@
...
+You can install Node.js v14.x by running the following commands:
+
+```
+$ curl -sL https://deb.nodesource.com/setup_13.x | sudo -E bash -
+$ sudo apt-get install -y nodejs
+```
+
+For other versions of Node.js refer to the [Github Page](ttps://github.com/nodesource/distributions/blob/master/README.md).
+
+Node js also installs npm which a package manager. You can update the version of npm using:
```
afustePTC, line 53: No need to have this here. Npm will be installed with node and the user doesn't need to update at this point, they could be confused so I would just remove this part.

RaspberryPi_README.md
```diff
@@ -0,0 +1,99 @@
...
+```
+$ npm run test
+```
+
+### Support
+If you encounter any problems you can make a post in the [Vuforia Spatial Toolbox Forum](https://forum.spatialtoolbox.vuforia.com/).
+
+###### Author: Saina Rezvani
+###### Date: 05/21/2020
+
```
afustePTC, line 98: Awesome!!

RaspberryPi_README.md
SainaRez: Should I take out the link to Node js github page as well?

afustePTC: No, it's ok you can leave it.
Just remove the npm update part. :)

PR #186, opened 2020-06-26
Anchor heartbeat bugfix and jpeg support

ddgond:
Fixed an issue where attaching a target to an anchor would make the server advertise both the anchor-version and target-version of the object until the server restarts. JPEG support was added to allow iOS to upload images from the camera roll.

### DIFF
--- a/libraries/utilities.js
+++ b/libraries/utilities.js
@@ -276,6 +276,27 @@ var getObjectIdFromTargetOrObjectFile = function (folderName, objectsPath) {
 };
 exports.getObjectIdFromTargetOrObjectFile = getObjectIdFromTargetOrObjectFile;
 
+var getAnchorIdFromObjectFile = function (folderName, objectsPath) {
+
+    if (folderName === 'allTargetsPlaceholder') {
+        return 'allTargetsPlaceholder000000000000';
+    }
+
+    var jsonFile = objectsPath + '/' + folderName + '/object.json';
+
+    if (fs.existsSync(jsonFile)) {
+        let thisObject = JSON.parse(fs.readFileSync(jsonFile, 'utf8'));
+        if (thisObject.hasOwnProperty('objectId')) {
+            return thisObject.objectId;
+        } else {
+            return null;
+        }
+    } else {
+        return null;
+    }
+};
+exports.getAnchorIdFromObjectFile = getAnchorIdFromObjectFile;
+
 /**
  *
  * @param folderName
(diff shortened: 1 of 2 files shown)


libraries/utilities.js
```diff
@@ -276,6 +276,27 @@ var getObjectIdFromTargetOrObjectFile = function (folderName, objectsPath) {
...
+        if (thisObject.hasOwnProperty('objectId')) {
+            return thisObject.objectId;
+        } else {
+            return null;
+        }
+    } else {
+        return null;
+    }
+};
+exports.getAnchorIdFromObjectFile = getAnchorIdFromObjectFile;
```
benptc, line 298: This is a good solution but I really want to solve this at the root of the problem – the object.json file at this path shouldn't exist (should always be inside the .identity folder). I actually had no idea that that bug existed until you noticed it.

server.js
```diff
@@ -4484,6 +4499,10 @@ function objectWebServer() {
                     if (req.headers.type === 'targetUpload') {
                         console.log('targetUpload', req.params.id);
                         var fileExtension = getFileExtension(filename);
+                        
+                        if (fileExtension === 'jpeg') { // Needed for compatibility, .JPEG is equivalent to .JPG
+                            fileExtension = 'jpg';
+                        }
```
benptc, line 4505: Very good to have!

server.js
```diff
@@ -1061,7 +1068,7 @@ function objectBeatSender(PORT, thisId, thisIp, oneTimeOnly) {
     });
 
     if (!oneTimeOnly) {
-        setInterval(function () {
+        activeHeartbeats[thisId] = setInterval(function () {
```
benptc, line 1071: Good idea to keep track of these so we can cancel heartbeat intervals if needed

server.js
```diff
@@ -4439,6 +4450,10 @@ function objectWebServer() {
                     var tempFolderName2 = utilities.readObject(objectLookup, req.body.name);//req.body.name + thisMacAddress;
                     // remove object from tree
                     if (tempFolderName2 !== null) {
+                        if (activeHeartbeats[tempFolderName2]) {
+                            clearInterval(activeHeartbeats[tempFolderName2]);
+                            delete activeHeartbeats[tempFolderName2];
+                        }
```
benptc, line 4456: This is good. I guess in the past, heartbeats of deleted objects continued to be broadcast until we restarted the server, but your new code is the correct behavior.

server.js
```diff
@@ -4625,6 +4644,16 @@ function objectWebServer() {
...
                                     setAnchors();
+                                    
+                                    // Removes old heartbeat if it used to be an anchor
+                                    var oldObjectId = utilities.getAnchorIdFromObjectFile(req.params.id, objectsPath);
+                                    if (oldObjectId && oldObjectId != thisObjectId) {
+                                      console.log('removed old heartbeat for', oldObjectId);
+                                      clearInterval(activeHeartbeats[oldObjectId]);
+                                      delete activeHeartbeats[oldObjectId];
+                                      delete objects[oldObjectId];
+                                    }
```
benptc, line 4655: This won't be necessary if we fix the root problem around anchors' object.json and their ID changing, but good for now.

PR #191, opened 2020-07-06
Automatic pattern generator for instant targets

### LINKED ISSUE #144: automatic pattern generator for instant targets

ddgond:
Resolves #144


libraries/webInterface/gui/index.js
```diff
@@ -2150,4 +2162,90 @@ function addZipDownload(button, frameName) {
     });
 }
 
+function voronoiTarget(canvas, callback) {
```
benptc, line 2165: This part works great, I'm glad the pattern generator was able to be incorporated like this.

libraries/webInterface/gui/index.js
```diff
@@ -1935,13 +1936,24 @@ realityServer.toggleFullScreen = function (item) {
...
+        const iframeContents = `<div style="text-align: center;"><div style="background: url(${targetUrl}) no-repeat center; background-size: contain; height: 100%; width: 100%;"></div></div>`;
+        fetch(targetUrl).then((response) => {
+            if (response.ok) {
+                thisIframe.contentDocument.write(iframeContents);
+                thisIframe.contentDocument.close();  
+            } else {
+                setGeneratedTarget(item, () => {
+                    thisIframe.contentDocument.write(iframeContents);
+                    thisIframe.contentDocument.close();    
+                });
```
benptc, line 1954: I like how it will automatically generate the target if needed here!

libraries/webInterface/gui/index.js
```diff
@@ -1935,13 +1936,24 @@ realityServer.toggleFullScreen = function (item) {
...
-    let screenPort = realityServer.objects[item.id.slice('fullscreen'.length)].screenPort;
-    thisIframe.src = 'http://' + realityServer.states.ipAdress.interfaces[realityServer.states.ipAdress.activeInterface] + ':' + screenPort;
-
     let thisScreen = thisIframe;
     // if(item) thisScreen = item;
 
     if (!thisScreen.mozFullScreen && !document.webkitFullScreen) {
+        thisIframe.src = "about:blank"; // Clear iframe before loading
+        const targetUrl = `/obj/${item.dataset.objectName}/target/target.jpg`;
+        const iframeContents = `<div style="text-align: center;"><div style="background: url(${targetUrl}) no-repeat center; background-size: contain; height: 100%; width: 100%;"></div></div>`;
```
benptc, line 1945: We should talk about the intended behavior before merging. This button was originally linked to a special hardware interface that I built which let you push and pull AR content in/out of screens. Opening the iframe whose src was set to the object's "screenPort" would open a webpage hosted at that endpoint which provided the capabilities for doing so. I need to fix a few things with that feature before including it again. For now, just showing the image is ok, but we will have to change it sometime in the future.

ddgond:
> Overall this is a really cool feature! I'm sure our users on the forum will love being able to just click generate.
> 
> What do you think of the following UX adjustments? We could add them in a follow-up PR:
> 
> * Should we have a way to delete a target from an object, to convert it back to an anchor? (e.g. if you accidentally clicked "generate" but actually want to an object to act as an anchor)
> * Should the generate button only exist for objects without targets already uploaded to them? It's currently possible to lose a target image that you previously uploaded if you click the generate button.

I do think having a way to remove targets and revert objects to anchors would be a useful addition, and if we add that, I think displaying the generate button only for anchors would be a useful ui change.

benptc:
Closing this because it was rebased into PR #203

Issue #195, opened 2020-07-07
Adding Hardware Nodes to Anchors does not work


valentinptc:
The nodes only show up, once the anchor has a target (therefore is not an anchor anymore)

PR #197, opened 2020-07-08
Hardware node anchor bugfix

### LINKED ISSUE #184: Anchors and non-anchors store object.json in different locations
Anchors store data at spatialToolbox/object.json and non-anchors store data at spatialToolbox/.identity/object.json. This becomes an issue when creating a non-anchor, since the first step requires creating it as an anchor before it gets a target applied to it, causing the server to read from both locations and advertise heartbeats for both objects.

ddgond:
Resolved #184 #195 
Anchors now store their data in `.identity/object.json` instead of `object.json`.
addNode can now add nodes to anchors (was a bug caused by above issue) and can now create anchors immediately if one doesn't exist (previously it simply created an empty folder and returned without the node).

### DIFF
--- a/libraries/hardwareInterfaces.js
+++ b/libraries/hardwareInterfaces.js
@@ -18,10 +18,12 @@
  */
 
 var path = require('path');
+var fs = require('fs');
 var utilities = require('./utilities');
 var _ = require('lodash');
 const Node = require('../models/Node.js');
 const Frame = require('../models/Frame.js');
+const ObjectModel = require('../models/ObjectModel.js');
 
 //global variables, passed through from server.js
 var objects = {};
@@ -34,6 +36,10 @@ var objectsPath;
 var nodeTypeModules;
 // eslint-disable-next-line no-unused-vars
 var blockModules;
+var services;
+var version;
+var protocol;
+var serverPort;
 var callback;
 var actionCallback;
 var publicDataCallBack;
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -820,7 +820,7 @@ function loadAnchor(anchorName) {
     // create the file for it if necessary
     var folder = path.join(objectsPath, anchorName);
     var identityPath = path.join(folder, '.identity');
-    var jsonFilePath = path.join(folder, 'object.json');
+    var jsonFilePath = path.join(identityPath, 'object.json');
```
benptc, line 823: Good catch

libraries/hardwareInterfaces.js
```diff
@@ -417,14 +423,45 @@ exports.setTool = function (object, tool, newTool, dirName) {
...
+            fs.writeFileSync(jsonFilePath, JSON.stringify(objects[objectID], null, 4), function (err) {
+                if (err) {
+                    console.log('anchor object save error', err);
+                } else {
+                    // console.log('JSON saved to ' + jsonFilePath);
+                }
+            });
+        } else {
+            console.log('I am not allowed to save');
+        }
```
benptc, line 464: This looks good!

server.js
```diff
@@ -867,20 +867,20 @@ function loadAnchor(anchorName) {
...
     if (globalVariables.saveToDisk) {
-
-        fs.writeFile(jsonFilePath, JSON.stringify(objects[anchorUuid], null, 4), function (err) {
+        fs.writeFileSync(jsonFilePath, JSON.stringify(objects[anchorUuid], null, 4), function (err) {
             if (err) {
                 console.log('anchor object save error', err);
             } else {
-                //console.log('JSON saved to ' + jsonFilePath);
+                // console.log('JSON saved to ' + jsonFilePath);
+                objectBeatSender(beatPort, anchorUuid, objects[anchorUuid].ip);
```
benptc, line 876: Ah yes, this was missing.

PR #205, opened 2020-07-15
Fixes node initialization to correctly load in default properties of the node type

benptc:
Resolves issue where default export.properties set by the node types were being ignored, because nodes were using default constructor instead of being type-specific.


libraries/utilities.js
```diff
@@ -875,4 +877,23 @@ const getVideoDir = function (objectsPath, identityFolderName, isMobile, objectN
...
 }
-exports.getVideoDir = getVideoDir;
\ No newline at end of file
+exports.getVideoDir = getVideoDir;
+
+exports.setupNodeTypes = function(nodeTypeModules) {
+    nodeTypes = nodeTypeModules;
+};
+
+const createNode = function(name, type) {
```
hobinjk-ptc, line 886: Why not have this be the Node constructor? It looks like every use of `new Node` has a name and type assigned at some point following. Is it for the private/public data?

benptc: Oh, that actually makes a lot of sense!

benptc: Essentially we just need the public and private data to be copied from the definition of the node type, so I think that's a reasonable modification of the constructor

libraries/availableModules.js
```diff
@@ -0,0 +1,33 @@
+let nodeTypes = {};
+let hardwareInterfaces = {};
+let blockTypes = {};
+
+const setNodes = function(nodeTypeModules) {
```
hobinjk-ptc, line 5: These declarations could be less redundant as
```javascript
exports.setNodes = function setNodes(nodeTypeModules) {
}
```

Technically the function name is also redundant and unnecessary but I think it might help in stack traces

models/Node.js
```diff
@@ -1,14 +1,15 @@
...
  * Constructor used to define every nodes generated in the Object. It does not need to contain its own ID
  * since the object is created within the nodes with the ID as object name.
  *
  * @constructor
  */
-function Node() {
+function Node(name, type) {
     // the name of each link. It is used in the Reality Editor to show the IO name.
-    this.name = '';
+    this.name = name;
```
hobinjk-ptc, line 12: Might be good for this to be `name || ''` in case something's doing a `name === ''` or equivalent

PR #216, opened 2020-07-30
Realtime settings to server from gui + fix redirection of public folder

afustePTC:
Push realtime settings to server from gui 
fix redirection of public folder files

### DIFF
--- a/server.js
+++ b/server.js
@@ -1555,6 +1555,7 @@ function objectWebServer() {
     // TODO: is the developer flag ever not true anymore? is it still useful to have?
     if (globalVariables.developer === true) {
         webServer.use('/libraries', express.static(__dirname + '/libraries/webInterface/'));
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1956,6 +1957,7 @@ function objectWebServer() {
                         } else {
                             console.log('successfully wrote settings hardwareInterface: ' + interfaceName);
                             callback(true);
+                            hardwareAPI.pushSettingsToGui(interfaceName, existingSettings);
```
hobinjk-ptc, line 1960: For safety it would be a good idea to add a mock of pushSettingsToGui to the mobile hardwareAPI object https://github.com/ptcrealitylab/vuforia-spatial-edge-server/blob/master/libraries/mobile/hardwareInterfaces.js

```
pushSettingsToGui: function() {
}
```

benptc: Oh, I forgot about that. Good catch

PR #222, opened 2020-08-12
hotfix for nodes with public data

benptc:
public data was being copied by reference (not by value) into each new node so they all shared the same publicData object on the server. so updating the public data of one node was updating them all.

E.g. when you added a drawing tool it would start out with the same drawing as the previously added drawing tools.

### DIFF
--- a/models/Node.js
+++ b/models/Node.js
@@ -43,8 +43,8 @@ function Node(name, type, objectId, frameId, nodeId) {
         this.publicData = {};
     } else {
         let nodeTemplate = nodeTypes[type];
-        this.privateData = nodeTemplate.properties.privateData || {};
-        this.publicData = nodeTemplate.properties.publicData || {};
+        this.privateData = Object.assign({}, nodeTemplate.properties.privateData) || {};
+        this.publicData = Object.assign({}, nodeTemplate.properties.publicData) || {};
     }
 
     this.setupProgram();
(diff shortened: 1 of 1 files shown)


models/Node.js
```diff
@@ -43,8 +43,8 @@ function Node(name, type, objectId, frameId, nodeId) {
         this.publicData = {};
     } else {
         let nodeTemplate = nodeTypes[type];
-        this.privateData = nodeTemplate.properties.privateData || {};
-        this.publicData = nodeTemplate.properties.publicData || {};
+        this.privateData = Object.assign({}, nodeTemplate.properties.privateData) || {};
```
hobinjk-ptc, line 46: publicData's members aren't guaranteed to be primitives as far as I know so this should use a deep copy like `JSON.parse(JSON.stringify(privateData))` instead (it might be nice to introduce a utilities.deepCopy which does a recursive Object.assign but that feels like a bit of a waste of time)

benptc: Thanks, good catch

models/Node.js
```diff
@@ -43,8 +43,8 @@ function Node(name, type, objectId, frameId, nodeId) {
         this.publicData = {};
     } else {
         let nodeTemplate = nodeTypes[type];
-        this.privateData = nodeTemplate.properties.privateData || {};
-        this.publicData = nodeTemplate.properties.publicData || {};
+        this.privateData = JSON.parse(JSON.stringify(nodeTemplate.properties.privateData)) || {};
```
hobinjk-ptc, line 46: The || should be inside the calls to stringify since JSON.parse(JSON.stringify(undefined)) is a syntax error

PR #254, opened 2020-12-01
allow Server to process complex Data Types

valentinptc:
The nodes are now able to process objects and not just numbers.
We need to test this a lot before we commit this change into the master.

There are corresponding nodes and blocks in the core add-on.
The system should run normal, however if objects propagate trough the system, the old core add-on might produce errors.

Next step: Making sure that the changes to the core add-on are backward compatible.


benptc (approved):
I've been testing this for awhile and it all seems to be working! 👍

PR #256, opened 2020-12-02
Update gl worker

hobinjk-ptc:
Brings in the latest and greatest from the standalone project

### DIFF
--- a/libraries/objectDefaultFiles/gl-worker.js
+++ b/libraries/objectDefaultFiles/gl-worker.js
@@ -54,6 +54,21 @@ function makeStub(functionName) {
       }
     }
 
+    if (functionName === 'texImage2D') {
+      let img = args[args.length - 1];
+      if (img.tagName) {
+        let width = img.width;
+        let height = img.height;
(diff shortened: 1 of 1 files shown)


libraries/objectDefaultFiles/gl-worker.js
```diff
@@ -51,24 +54,43 @@ function makeStub(functionName) {
       }
     }
 
+    if (functionName === 'texImage2D') {
```
benptc, line 57: just curious, do you anticipate many cases that need to be handled individually like this, or is texImage2D an outlier?

benptc (approved):
great work! looks good to me

libraries/objectDefaultFiles/gl-worker.js
hobinjk-ptc: It's the image part of texImage2D that causes an issue so I think it's an outlier. There's plenty of room for things to require special-casing in other ways (and I had to modify a tool's three.min.js file to work around one)

PR #288, opened 2021-02-24
Updates for screens addon

benptc:
- Updates web interface with tooltip help UI
- Fixes tool repositioning API so that scene graph doesn't use incorrect position data from tools inside of screens
- Adds a few hardwareInterface APIs that the screens interface can use to initialize new objects correctly, set their target size, and switch their visualization mode from AR to screen.
- Adds some error handling for target.xml files with invalid contents.

### DIFF
--- a/controllers/object.js
+++ b/controllers/object.js
@@ -196,10 +196,14 @@ const activate = function(objectID, callback) {
 };
 
 const setVisualization = function(objectID, vis, callback) {
-    utilities.getObject(objects, objectID).visualization = vis;
-    console.log(vis, objectID);
-    utilities.writeObjectToFile(objects, objectID, objectsPath, globalVariables.saveToDisk);
-    callback(200, 'ok');
+    try {
+        utilities.getObject(objects, objectID).visualization = vis;
+        console.log(vis, objectID);
+        utilities.writeObjectToFile(objects, objectID, objectsPath, globalVariables.saveToDisk);
+        callback(200, 'ok');
+    } catch (e) {
+        callback(404, {success: false, error: 'cannot find object with ID' + objectID});
+    }
 };
 
 // request a zip-file with the object stored inside
--- a/controllers/object.js
+++ b/controllers/object.js
@@ -196,13 +196,16 @@ const activate = function(objectID, callback) {
 };
 
 const setVisualization = function(objectID, vis, callback) {
+    let object = utilities.getObject(objects, objectID);
+    if (!object) {
+        callback(404, {success: false, error: 'cannot find object with ID' + objectID});
+    }
     try {
-        utilities.getObject(objects, objectID).visualization = vis;
-        console.log(vis, objectID);
+        object.visualization = vis;
         utilities.writeObjectToFile(objects, objectID, objectsPath, globalVariables.saveToDisk);
         callback(200, 'ok');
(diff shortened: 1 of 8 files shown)


controllers/object.js
```diff
@@ -196,10 +196,14 @@ const activate = function(objectID, callback) {
 };
 
 const setVisualization = function(objectID, vis, callback) {
-    utilities.getObject(objects, objectID).visualization = vis;
-    console.log(vis, objectID);
-    utilities.writeObjectToFile(objects, objectID, objectsPath, globalVariables.saveToDisk);
-    callback(200, 'ok');
+    try {
```
hobinjk-ptc, line 199: It would be a bit better to write this as
```
let object = utilities.getObject(...);
if (!object) {
  callback with 404
  return
}
rest of function
```
since right now if writeObjectToFile throws an error (like if the disk is full or there aren't permissions for objectsPath) it will be misreported as a 404

libraries/hardwareInterfaces.js
```diff
@@ -694,14 +699,62 @@ exports.deactivate = function (object) {
...
     return utilities.getObjectIdFromTargetOrObjectFile(object, objectsPath);
 };
 
 exports.getMarkerSize = function (object) {
+    try {
+        var objectID = utilities.getObjectIdFromTargetOrObjectFile(object, objectsPath);
+        return objects[objectID].targetSize;
+    } catch (e) {
+        console.warn('Cannot get markerSize for ' + object + ', returning (0,0)');
+        return { width: 0, height: 0};
```
hobinjk-ptc, line 723: nit: inconsistent spacing around braces

libraries/webInterface/gui/index.js
```diff
@@ -262,6 +388,33 @@ realityServer.updateManageObjects = function (thisItem2) {
                 thisObject.dom.querySelector('.target').setAttribute('isWorldObject', true);
                 thisObject.dom.querySelector('.target').addEventListener('click', realityServer.gotClick, false);
 
+                setTooltipTextForElement('.name',
```
hobinjk-ptc, line 391: There's multiple things with class name so the setTooltipTextForElement function should use querySelectorAll to give them all tooltips

libraries/webInterface/gui/index.js
```diff
@@ -436,6 +643,9 @@ realityServer.updateManageObjects = function (thisItem2) {
                     realityServer.switchClass(thisObject.dom.querySelector('.target'), 'green', 'yellow');
                     realityServer.switchClass(thisObject.dom.querySelector('.target'), 'one', 'targetWidthMedium');
 
+                    setTooltipTextForElement(thisObject.dom.querySelector('.target'), 'Add a target file or activate a world object to finish' +
```
hobinjk-ptc, line 646: It feels like this doing the querySelector manually is equivalent to what the optionalParent argument accomplishes by providing a different querySelector target

hobinjk-ptc (approved):
Super nice to have all these improvements!

libraries/webInterface/gui/index.js
benptc: Yeah I didn't expect to need to use it as often as I did, I'll just go back and change them to standard querySelector syntax

PR #301, opened 2021-03-13
Complext link trigger

valentinptc:
This makes sure that once a link is created for a complex data node, the data node data is traveling to the destination right after. Its only processing this specific link for the data node.


hobinjk-ptc (approved):
LGTM

hobinjk-ptc:
but I agree that @benptc is the better person to review this :D

benptc (approved):
Looks good! I tested it, and this seems to work!

PR #303, opened 2021-03-17
adding gl-worker.js, envelope.js and envelopeContents.js path for custom tools

afustePTC:


hobinjk-ptc (approved):
Good fix for now! Would be nice to make this a bit less repetitive in the future

benptc (approved):
Yep, this looks good!

benptc:
> Good fix for now! Would be nice to make this a bit less repetitive in the future

@hobinjk-ptc Agreed. One way would just be to iterate through the *.js filenames in the objectDefaultFiles directory on the server, and perform this replacement for each match. Do you know of a better way?

hobinjk-ptc:
Oh that would be reasonable. I was thinking more on how this duplicates code from /frames/:frameName

benptc:
Oh right, that too! Seems like there would be an easy enough way to refactor them 👍

PR #310, opened 2021-04-16
Glb upload

dangond-ptc:
Added the ability to upload glb files from area targets to enable navigation features in the toolbox.

### DIFF
--- a/server.js
+++ b/server.js
@@ -324,6 +324,11 @@ var formidable = require('formidable'); // Multiple file upload library
 var cheerio = require('cheerio');
 var request = require('request');
 
+// use the cors cross origin REST model
+webServer.use(cors());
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -326,6 +326,11 @@ var formidable = require('formidable'); // Multiple file upload library
 var cheerio = require('cheerio');
 var request = require('request');
 
+// use the cors cross origin REST model
+webServer.use(cors());
```
hobinjk-ptc, line 330: The old webServer.use(cors()) later on in this file should be deleted so it isn't redundant

dangond-ptc: Good catch, I had originally moved it up to the new position but I assume a merge put it back at the original spot at some point

benptc (approved):
Tested this and it all seems to work!

PR #315, opened 2021-05-05
Pinned tools, upload media files from tools, and support for addon content_styles and content_resources

benptc:
Adds `isPinned` property to tools, adds optional `excludeUnpinned` queryParam when getting object data to ignore unpinned tools, and adds a `GET /spatial/search` API to discover/download tools (e.g. unpinned tools) based on proximity (and other filters).

Also adds a new `POST /:objectName/uploadMediaFile` for uploading photos and videos (made for the communication tool but can be used for more things in future)

Addons can now also have a `content_styles` directory with CSS files and a content_resources directory with SVGs and PNGs that can be accessed from the content_scripts


controllers/object.js
```diff
@@ -63,6 +63,58 @@ const uploadVideo = function(objectID, videoID, reqForForm, callback) {
...
+    var form = new formidable.IncomingForm({
+        uploadDir: mediaDir,
+        keepExtensions: true
+        // accept: 'image/jpeg' // TODO: specify which types of images/videos it accepts?
+    });
+
+    console.log('created form');
+
+    form.on('error', function (err) {
+        callback(500, err);
```
dangond-ptc, line 89: We probably want a return statement here, right?

dangond-ptc (approved):

controllers/object.js
```diff
@@ -63,6 +63,58 @@ const uploadVideo = function(objectID, videoID, reqForForm, callback) {
...
+    if (fs.existsSync(rawFilepath)) {
+        console.log('deleted old raw file');
+        fs.unlinkSync(rawFilepath);
+    }
+
+    form.on('fileBegin', function (name, file) {
+        console.log('fileBegin loading', name, file);
+
+        if (file.type.match('video.*')) {
+            rawFilepath = rawFilepath.replace('.jpg', '.mov');
```
hobinjk-ptc, line 105: Is there a less fragile way to do this? Or is this working around a specific jpg <-> mov bug?

hobinjk-ptc (approved):

controllers/object.js
benptc: Yeah I can do this in a less fragile way

PR #329, opened 2021-05-07
Object.type, human objects, batched realtime updates, etc

benptc:
- adds .type to objects that supports object|world|human
- adds new /batchedUpdate socket message listener for more efficient realtime x,y,scale,etc updates
- clicking on world object name in localhost:8080 will open the remote operator if that add-on is set up
- removes some now-unnecessary logs
- fixes a couple tiny bugs

### DIFF
--- a/libraries/utilities.js
+++ b/libraries/utilities.js
@@ -63,7 +63,6 @@ var path = require('path');
 var request = require('request');
 const fetch = require('node-fetch');
 const ObjectModel = require('../models/ObjectModel.js');
-const URLSearchParams = require('url').URLSearchParams;
 
 var hardwareInterfaces = {};
 
--- a/libraries/utilities.js
+++ b/libraries/utilities.js
@@ -63,6 +63,7 @@ var path = require('path');
 var request = require('request');
 const fetch = require('node-fetch');
 const ObjectModel = require('../models/ObjectModel.js');
+const URLSearchParams = require('url').URLSearchParams;
 
 var hardwareInterfaces = {};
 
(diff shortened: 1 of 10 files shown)


models/ObjectModel.js
```diff
@@ -55,7 +55,9 @@ function ObjectModel(ip, version, protocol, objectId) {
         width: 0.3, // default size should always be overridden, but exists in case xml doesn't contain size
         height: 0.3
     };
-    this.isWorldObject = false;
+    this.isWorldObject = false; // a bit redundant with this.type, but good for backwards compatibility
```
hobinjk-ptc, line 58: This is a bit overly fancy but if you rewrite this as a getter/setter pair (<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get>) you won't have to worry about it getting out of sync with this.type

server.js
```diff
@@ -3403,11 +3402,30 @@ function socketServer() {
 
         });
 
+        // relays realtime updates (to matrix, x, y, scale, etc) from one client to the rest of the clients
+        // clients are responsible for batching and processing the batched updates at whatever frequency they prefer
+        socket.on('/batchedUpdate', function (msg) {
```
hobinjk-ptc, line 3407: Awesome API to have!

hobinjk-ptc (approved):

models/ObjectModel.js
hobinjk-ptc: (might only need a getter)

libraries/utilities.js
```diff
@@ -63,6 +63,7 @@ var path = require('path');
 var request = require('request');
 const fetch = require('node-fetch');
 const ObjectModel = require('../models/ObjectModel.js');
+const URLSearchParams = require('url').URLSearchParams;
```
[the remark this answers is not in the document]
benptc: Ah ok – I ran into a bug earlier but good to know it's fixed now without this

PR #359, opened 2021-06-28
Pass width and height from bootstrap into main

hobinjk-ptc:


benptc (approved):
Tested this out on one tool, seems to work!

PR #371, opened 2021-08-02
small stability improvements

benptc:
- fixes corrupted heartbeat checksums when server restarts
- catches error if there's an old server running on the network that doesn't support scenegraph APIs
-  adds a param when clicking on remote operator link (does nothing for now, but userinterface can make use of it later)


dangond-ptc (commented):
Looks good to me!

dangond-ptc (approved):
Forgot to hit approve

benptc:
> Looks good to me!

Thanks!

PR #387, opened 2021-08-31
Make .identity folders optional for frames; Default to enabled

hobinjk-ptc:


benptc (approved):
Thanks for this! I've run into this issue before

PR #421, opened 2021-11-04
adds a POST /object/:id/generateFrame route 

benptc:
How does this look?

The request just needs frameType and relativeMatrix in the body and does all the other work to create the frame on the specified object and notify any clients to load it.

example: 

```
curl --header "Content-type: application/json" -d "{\"frameType\":\"graphUI\", \"relativeMatrix\":[1,0,0,0,0,1,0,0,0,0,1,0,0,0,0,1]}" http://localhost:8080/object/test_Peoqk0h4lgb/generateFrame 
```
^ This will create a graphUI frame and add it to my object named test (uuid = test_Peoqk0h4lgb), and position it centered on the object

### DIFF
--- a/controllers/frame.js
+++ b/controllers/frame.js
@@ -79,6 +79,40 @@ const addFrameToObject = function (objectKey, frameKey, frame, callback) {
     });
 };
 
+const generateFrameOnObject = function (objectKey, frameType, relativeMatrix, callback) {
+    let object = utilities.getObject(objects, objectKey);
+    if (!object) {
+        callback(404, error);
+        return;
+    }
+
+    if (!object.frames) {
+        object.frames = {};
+    }
+
+    let frameKey = objectKey + frameType + utilities.uuidTime();
+    let newFrame = new Frame(objectKey, frameKey);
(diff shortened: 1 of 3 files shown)


benptc:
Here's a sample for an add-on interface to add a new 3D sphere tool to the first world object on the server every 5 seconds. If you have the remote operator open while this runs, you should see them pop up as they're added:

```javascript
setInterval(function() {
    let objects = server.getAllObjects();
    let worldId = Object.keys(objects).find(function(key) {
        let isWorld = objects[key].isWorldObject || objects[key].type === 'world';
        return isWorld && objects[key].deactivated === false;
    });
    if (worldId) {
        let x = Math.random() * 500 - 1000;
        let y = Math.random() * 500 - 1000;
        let z = Math.random() * 500 - 1000;
        let newFrame = server.generateFrame(objects[worldId].name, 'sphere', [1,0,0,0,0,1,0,0,0,0,1,0,x,y,z,1]);
    }
}, 5000);
```

dangond-ptc (approved):
Looks good to me!

PR #430, opened 2021-11-12
fix longtime subtle bug where fullscreen tools (e.g. threejs) blocked pointermove and pointerup...

benptc:
...events on tools underneath them even if touchDecider (raycast) returned false within the tool. Noticed this behavior on the remote operator where communication tools would received pointerdown events when threejs tools are in the scene, but not pointerup or pointermove. Adjusted object.js library to propagate these events too.


hobinjk-ptc (approved):
Awesome fix to have!

dangond-ptc (approved):
Looks good to me!

PR #442, opened 2021-12-02
adds support for optional unitypackage target file (in addition to xml, dat, and glb)

benptc:
New iOS app version will generate this file from the area target scanner and include it in the zip that is uploaded to the server. `target.unitypackage` is available at path, e.g. http://localhost:8080/obj/_WORLD_instantScan/target/target.unitypackage


afustePTC (approved):
Thanks!

PR #444, opened 2021-12-06
Glproxy threejs interface

hobinjk-ptc:


benptc (approved):
This looks great! Really awesome to have. I checked out the pr too and it all seems to work. This'll make it so much simpler to build more 3D tools in future.

PR #453, opened 2021-12-22
Always setup LocalUIApp; Serve fbx, gltf, and glb files

hobinjk-ptc:
The decision to only LocalUIApp looks like it's out of sync with LocalUIApp also serving add-on resources.

After this it'll be possible to access /addons/vuforia-spatial-remote-operator-addon/content_resources/hololens.fbx at http://localhost:49368/addons/vuforia-spatial-remote-operator-addon/hololens.fbx


benptc (approved):
LGTM

PR #476, opened 2022-02-08
Add file.name as alias to newFilename

hobinjk-ptc:


hobinjk-ptc:
Tested locally and it didn't work the first time but worked after that 🤷

dangond-ptc (approved):
If this works for you we can push it, I've been testing on my end and there's several other things on top of this that I need to modify before I can get it working.

hobinjk-ptc:
Alright, sounds good

PR #478, opened 2022-02-10
Agent support merge

hobinjk-ptc:
Main highlights are changes in object.js and server.js that amount to hooking up Toolsocket to both ends of communication. Unfortunately this also picked up some eslint fixes


libraries/objectDefaultFiles/object.js
```diff
@@ -101,6 +100,29 @@
     // this will be initialized once the frame creates a new SpatialInterface()
     var realityInterface = null;
 
+    let urlObj = {};
+
+    function getIoTitle (title) {
```
benptc, line 105: This seems to be used a lot, but I can't quite decipher what it's doing (what is urlObj.n and when is it defined or not?) – do you mind adding a comment for future readability?

server.js
```diff
@@ -3260,11 +3294,22 @@ function socketServer() {
...
+                        isNew = false;
+                    }
+                });
+
+                if (isNew) {
+                    realityEditorSocketArray[socket.id].push({
+                        object: msgContent.object,
+                        frame: msgContent.frame,
+                        protocol: thisProtocol
+                    });
```
benptc, line 3311: I see we're now adding the subscriptions to an array rather than only having one per socket – does this fix any bugs we were having / change any functionality? Seems like the correct approach but I'm wondering why we didn't do this before

server.js
```diff
@@ -3770,12 +3852,15 @@ function sendMessagetoEditors(msgContent, sourceSocketID) {
...
 
     for (var thisEditor in realityEditorSocketArray) {
-        if (typeof sourceSocketID !== 'undefined' && thisEditor === sourceSocketID) {
-            continue; // don't trigger the read listener of the socket that originally wrote the data
-        }
-        if (msgContent.object === realityEditorSocketArray[thisEditor].object && msgContent.frame === realityEditorSocketArray[thisEditor].frame) {
-            messagetoSend(msgContent, thisEditor);
-        }
+        realityEditorSocketArray[thisEditor].forEach((thisObj) => {
+            if (typeof sourceSocketID !== 'undefined' && thisEditor === sourceSocketID && msgContent.object === thisObj.object && msgContent.frame === thisObj.frame) {
```
benptc, line 3856: I see this checks more strictly (as opposed to previous version) that the subscription ignored is the exact object + frame that sent it – is that fixing an existing bug or required for the design change. Just wrapping my head around how I should write socket-handling code in the future.

server.js
```diff
@@ -3853,9 +3938,10 @@ function messagetoSend(msgContent, socketID) {
 
     var node = getNode(msgContent.object, msgContent.frame, msgContent.node);
     if (node) {
+        // console.log(socketID);
         io.sockets.connected[socketID].emit('object', JSON.stringify({
             object: msgContent.object,
-            frames: msgContent.frame,
+            frame: msgContent.frame,
```
benptc, line 3944: yikes, good typo catch. how was this working before? 😆

benptc (approved):
Looks good! Thanks for all the nitty gritty changes to make it work. Added a few comments where I was confused, but you can definitely merge it and I can resolve my confusion later.

server.js
hobinjk-ptc: I'm not entirely sure. It might be working around a bug where toolsockets from the socket.io compatibility server don't appear to get unique ids but I honestly don't know. I agree that it seems like the more correct approach

server.js
hobinjk-ptc: This is another one that I'm not too sure about. I believe this is the source of the `frames` typo below becoming a bug

libraries/objectDefaultFiles/object.js
hobinjk-ptc: Added some comments, thanks for pointing this out!

PR #518, opened 2022-04-20
Standalone Server Support

hobinjk-ptc:
Mostly enables some features using isStandaloneMobile as a replacement for isMobile. Also has support for 10.5/10.6 area target files

### DIFF
--- a/server.js
+++ b/server.js
@@ -77,8 +77,8 @@ try {
 const _logger = require('./logger');
 
 const os = require('os');
-const isMobile = false; // os.platform() === 'android' || os.platform() === 'ios' || process.env.FORCE_MOBILE;
-const isCoolMobile = true;
+const isMobile = os.platform() === 'android' || process.env.FORCE_MOBILE;
+const isCoolMobile = os.platform() === 'ios';
 
 // These variables are used for global status, such as if the server sends debugging messages and if the developer
 // user interfaces should be accesable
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -77,7 +77,8 @@ try {
 const _logger = require('./logger');
 
 const os = require('os');
-const isMobile = os.platform() === 'android' || os.platform() === 'ios' || process.env.FORCE_MOBILE;
+const isMobile = os.platform() === 'android' || process.env.FORCE_MOBILE;
```
benptc, line 80: Would it make sense to rename isMobile to something more accurate, since an iOS device is no longer considered mobile here? I can just foresee myself potentially misusing isMobile in the future.

server.js
```diff
@@ -1112,7 +1114,8 @@ function objectBeatSender(PORT, thisId, thisIp, oneTimeOnly) {
         let datPath = path.join(targetDir, 'target.dat');
         let xmlPath = path.join(targetDir, 'target.xml');
         let glbPath = path.join(targetDir, 'target.glb');
-        var fileList = [jpgPath, xmlPath, datPath, glbPath];
+        let tdtPath = path.join(targetDir, 'target.3dt');
```
benptc, line 1117: great to support this!

benptc (approved):
Thanks for the thorough and detail-oriented updates! I have one suggestion regarding the variable naming but feel free to merge the PR.

server.js
hobinjk-ptc: Agreed! I've renamed it and simplified some logic later on in the file that had remnants of the past setup where isMobile had a different meaning

benptc: Awesome!

PR #527, opened 2022-05-05
adds a projectedZ property to touch events, if the userinterface provides it

benptc:
Tools can access the projectedZ from pointerevent listeners and also in the touchDecider function

e.g.
```
document.addEventListener('pointerdown', e => {
  console.log(e.projectedZ);
})
```
or
```
spatialInterface.registerTouchDecider(e => {
  console.log(e.projectedZ);
  // ...
  return intersects.length > 0;
})
```


dangond-ptc (approved):

benptc:
Fixing a bug in the corresponding userinterface change to support this feature - I'll make a PR for that part soon

hobinjk-ptc (approved):

PR #555, opened 2022-07-12
Create and delete avatar objects

benptc:
Avatar objects can be created by the userinterface. They are deleted by the server when 1) a client disconnects (detected by socket disconnect), or 2) all avatar objects are deleted when the server restarts

### DIFF
--- a/server.js
+++ b/server.js
@@ -2420,6 +2420,11 @@ function objectWebServer() {
                             objectType = 'avatar';
                         }
                         objects[objectId].type = objectType;
+
+                        if (typeof req.body.worldId !== 'undefined') {
+                            objects[objectId].worldId = req.body.worldId;
+                        }
+
                         utilities.writeObjectToFile(objects, objectId, objectsPath, globalVariables.saveToDisk);
                         utilities.writeObject(objectLookup, req.body.name, objectId);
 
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1014,6 +1014,31 @@ function setAnchors() {
...
+    }
+
+    // populate all objects folders with object.json files.
+    tempFiles.forEach(function (objectKey) {
+        if (objectKey.includes('_AVATAR_')) {
+            objectsToDelete.push(objectKey);
+        }
+    });
+    
+    objectsToDelete.forEach(objectKey => {
```
hobinjk-ptc, line 1036: Based on this using rmSync over the deleteObjects API it feels like this would be a bit clearer named removeAvatarFiles

server.js
```diff
@@ -1014,6 +1014,31 @@ function setAnchors() {
...
+    // load all object folders
+    let tempFiles = fs.readdirSync(objectsPath).filter(function (file) {
+        return fs.statSync(path.join(objectsPath, file)).isDirectory();
+    });
+    // remove hidden directories
+    while (tempFiles.length > 0 && tempFiles[0][0] === '.') {
+        tempFiles.splice(0, 1);
+    }
+
+    // populate all objects folders with object.json files.
```
hobinjk-ptc, line 1029: nit: Comment is incorrect

hobinjk-ptc (approved):

PR #562, opened 2022-07-26
properly delete the avatar objects from the server when it restarts (…

benptc:
…don't just delete folder but also completely delete from the objects data structure)

There was a bug where restarting the server while a remote operator client is still open would leave it's avatar object in the data structure, and its `.type` would be corrupted from `avatar` to `object`, making it visible to clients. These problems are now avoided by deleting the objects before the server starts to use them.

### DIFF
--- a/server.js
+++ b/server.js
@@ -788,6 +788,10 @@ function loadObjects() {
         }
     }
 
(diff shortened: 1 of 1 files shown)


dangond-ptc (approved):
Looks good to me!

PR #569, opened 2022-08-09
Support WebGL2; Proxy VAO extension methods using built-in WebGL2 VAO…

hobinjk-ptc:
… methods

@dangond-ptc It might be worth checking to see how this impacts spatialDraw performance in your opinion. I thought it was mostly the same but I also noticed some degradation when clicking to move the camera where I wasn't sure of the cause.

### DIFF
--- a/libraries/objectDefaultFiles/gl-worker.js
+++ b/libraries/objectDefaultFiles/gl-worker.js
@@ -1,5 +1,5 @@
 const debugGlWorker = false;
-const GLPROXY_ENABLE_EXTVAO = false;
+const GLPROXY_ENABLE_EXTVAO = true;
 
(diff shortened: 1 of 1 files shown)


dangond-ptc:
Just tested both the current master branches and webgl2-support branches with and without the drawing tool, but couldn't see a difference in lag when moving the camera. In all cases, I had noticeable lag if not plugged in to power but it was roughly the same with and without the tool, and with and without the new changes. As far as I can tell, these commits did not make performance worse on my end.

dangond-ptc (approved):
Looks good to me!

hobinjk-ptc:
Great, thanks for the confirmation!

PR #576, opened 2022-08-16
Add support for internal /disconnectEditor message from cloud proxy

hobinjk-ptc:


benptc (approved):
Great, thanks for adding this!

PR #578, opened 2022-08-19
doesnt rely on a specific socket subscription when /disconnectEditor is received

benptc:
Fixes bug where the device's local server wouldn't delete its own avatar (when Exit Button is pressed in pop-up app)


hobinjk-ptc (approved):
A nice simplification!

PR #582, opened 2022-08-25
fix undefined thisObjectId, and standardize type='world' when isWorldObject=true

benptc:
trying to fix bug where occasionally instantScan worlds end up with .type='object' instead of 'world'. Fixed a more general mistake (perhaps causing side effects) where objects get re-created every time any target data is uploaded, rather than just the first time.

### DIFF
--- a/server.js
+++ b/server.js
@@ -851,6 +851,7 @@ function loadWorldObject() {
     worldObject.port = serverPort;
     worldObject.name = worldObjectName;
     worldObject.isWorldObject = true;
+    worldObject.type = 'world';
 
     // try to read previously saved data to overwrite the default world object
     if (globalVariables.saveToDisk) {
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -2901,11 +2902,13 @@ function objectWebServer() {
...
                                 // create the object data and respond to the webFrontend once the XML file is confirmed to exist
                                 function onXmlVerified(err) { // eslint-disable-line no-inner-declarations
+                                    let thisObjectId = utilities.readObject(objectLookup, req.params.id);
+
                                     if (err) {
                                         console.log(err);
                                     } else {
                                         // create the object if needed / possible
-                                        if (typeof objects[thisObjectId] === 'undefined') { // TODO: thisObjectId is always undefined?
+                                        if (typeof objects[thisObjectId] === 'undefined') {
```
hobinjk-ptc, line 2911: Good to have this frightening comment resolved!

hobinjk-ptc (approved):

PR #585, opened 2022-08-29
Mobile cleanup update

hobinjk-ptc:
Running scripts/mobile-cleanup.sh will result in a pretty significant improvement to compile and install times. Other wins can be had by running `npm ci --omit=dev` in all the addon folders and removing `@ffmpeg-installer/ffmpeg` from the remote operator addon.


benptc (approved):
Awesome, thank you for updating this!

dangond-ptc (approved):
Looks good to me!

PR #607, opened 2022-10-10
Integration test of server, remote operator, and edge agent

hobinjk-ptc:


test.js
```diff
@@ -4,10 +4,84 @@
...
+    });
+
+    await page.waitForSelector('#gltf-added', {
+        timeout: 60 * 1000
+    });
+
+    await page.screenshot({
+        path: 'screenshots/remote-operator-loaded-localhost.png',
+        fullPage: true,
+    });
```
benptc, line 55: Very cool to have these automated tests!

benptc (approved):
Awesome to have this! I wasn't sure what the best way to test things would be but I like your approach of taking screenshots along the way

hobinjk-ptc:
Thanks! I view this as really only the first step but my hope is to go from screenshots to comparing screenshots between revisions and finally to proper testing that API calls and user interactions are handled properly by the server

PR #613, opened 2022-10-17
adds spatialInterface.setAlwaysFaceCamera(true) API to toggle billboarding effect for 2D tools

benptc:
@dangond-ptc you can use this in the spatialDraw tool now. Fullscreen/stickyfullscreen takes precedent over this so you can leave it on and it won't affect the threejs scene within the tool, only when it's in the icon view.


dangond-ptc (approved):
Looks good to me!

PR #625, opened 2022-11-02
Start of objectsPath refactor

hobinjk-ptc:
In a perfect world, every injection of objectsPath (and other shared variables) is done by requiring a file so we don't have error-prone calls to setup(every, variable, in, server, dot, js) but it's not really a high priority to fix this considering the server currently works

### DIFF
--- a/config.js
+++ b/config.js
@@ -1,0 +1,38 @@
+/**
+ * Server configuration
+ *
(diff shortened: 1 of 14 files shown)


dangond-ptc (approved):
Looks good to me! Every refactoring of server.js soothes my soul

benptc (approved):
Much appreciated! LGTM

PR #626, opened 2022-11-02
Server heartbeat with services array from --services arg

benptc:


hobinjk-ptc:
This is missing an update to package-lock.json for yargs (probably just need to run `npm install`)

hobinjk-ptc (approved):

PR #629, opened 2022-11-04
Sends server UDP beat in response to client 'ping' messages

benptc:
Allows the clients to discover the server (and which services it has) more quickly, since they send a ping message when they first open.


dangond-ptc (approved):
Looks good to me!

hobinjk-ptc (approved):

PR #630, opened 2022-11-07
Drop forced nulling of context and domElement

hobinjk-ptc:
On some versions of three.js this raises an error and it shouldn't be necessary given that we're dropping the overall reference to realRenderer


ptc-rdeleeuw:
What kind of error does it generate?

hobinjk-ptc:
It says that `context` is a get-only property provided by WebGLRenderingContext

hobinjk-ptc:
@ptc-rdeleeuw Does this look good or will it conflict with your work in progress?

ptc-rdeleeuw:
The change seems to be cosmetic only, in removing the error. It doesn't seem to interfere with my work.

ptc-rdeleeuw (approved):

PR #634, opened 2022-11-08
Spatial video support

dangond-ptc:
Adds API support for tools using the RVL playback feature in the user interface.

### DIFF
--- a/libraries/objectDefaultFiles/gl-worker.js
+++ b/libraries/objectDefaultFiles/gl-worker.js
@@ -391,8 +391,8 @@ class ThreejsInterface {
 
     main({width, height}) {
         this.spatialInterface.changeFrameSize(width, height);
-        const canvas = document.createElement('canvas');
-        realGl = canvas.getContext('webgl2');
+        this.canvas = document.createElement('canvas');
+        realGl = this.canvas.getContext('webgl2');
         this.realRenderer = new THREE.WebGLRenderer({context: realGl, alpha: true});
         this.realRenderer.debug.checkShaderErrors = false;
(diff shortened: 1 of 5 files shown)


libraries/objectDefaultFiles/gl-worker.js
```diff
@@ -481,3 +485,43 @@ class ThreejsInterface {
...
+        this.preRender(now);
+
+        if (this.isProjectionMatrixSet) {
+            if (this.realRenderer && this.scene && this.camera) {
+                this.realRenderer.render(this.scene, this.camera);
+            }
+        }
+    }
+
+    setMatrixFromArray(matrix, array) {
```
hobinjk-ptc, line 512: This shouldn't need to exist since it'll naturally call the superclass' implementation

libraries/serverHelpers/toolboxEdgeProxyRequestHandler.js
```diff
@@ -0,0 +1,15 @@
+const https = require('https');
+
+const toolboxEdgeProxyRequestHandler = (req, res) => {
```
hobinjk-ptc, line 3: It would be good to have documentation on what req and res are expected to be and what format is expected from req's params

hobinjk-ptc (approved):
LGTM, excited to have this landed!

libraries/serverHelpers/toolboxEdgeProxyRequestHandler.js
dangond-ptc: These are the same params that Express passes on to its handlers, since it's being used as a handler on some Express routes

PR #645, opened 2022-11-30
Add humanPose and three to objectDefaultFiles; Lint fixes; Parser size increases

hobinjk-ptc:


benptc (approved):
Still feels a bit suboptimal to have all the code duplicated, but I don't mind doing this at least as a temporary solution.

PR #646, opened 2022-11-30
Add API docs from internal wiki

hobinjk-ptc:
I was thinking that it would be good to have this documentation public and next to its implementation (even though it's probably only useful for complete nerds).


benptc (approved):
Docs are always appreciated!

PR #663, opened 2023-01-10
Apply update messages to objects immediately

hobinjk-ptc:
Previous update and batchedUpdate users would always end their stream of updates with a separate API call to persist their modification (e.g. tool repositioning) but pose tracking doesn't have a nice end that's expected to happen regularly enough for this to make sense. Instead this writes directly to objects based on each property update which allows the objects variable to always be consistent with the user interface clients' state

### DIFF
--- a/server.js
+++ b/server.js
@@ -3691,8 +3691,63 @@ function socketServer() {
             }
         });
 
+        /**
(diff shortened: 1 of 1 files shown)


benptc (approved):
Looks good!

PR #673, opened 2023-01-24
History and logging

hobinjk-ptc:
Fixes up the existing recorder functionality to add periodic checkpoints and compression. This achieves a compression ratio of 0.2 with just gzip, the disabled compressFloat functionality adds another 0.8

### DIFF
--- a/libraries/recorder.js
+++ b/libraries/recorder.js
@@ -199,14 +199,17 @@ recorder.recurse = function (obj, objectInTime, keyString) {
             let thisValue = thisItem[key];
             let oldValue = oldItem[key];
 
-            if (typeof thisValue === 'number' && typeof oldValue === 'number') {
+            if (typeof thisValue === 'number') {
                 thisValue = compressFloat(thisValue);
+            }
+            if (typeof oldValue === 'number') {
                 oldValue = compressFloat(oldValue);
             }
 
-            if (thisItem[key] !== oldItem[key]) {
+            if (thisValue !== oldValue) {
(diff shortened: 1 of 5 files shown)


dangond-ptc (approved):
Looks good to me!

benptc (approved):
Looks good! Nice to have these history APIs available, too!

libraries/recorder.js
```diff
@@ -1,70 +1,190 @@
-var fs = require('fs');
+const fs = require('fs');
```
ptc-mklaudiny, line 1: It would be great for newcomers to have brief comments what main methods do.

routers/history.js
```diff
@@ -0,0 +1,83 @@
...
+        res.set('Content-Encoding', 'gzip');
+        const readStream = fs.createReadStream(compressedLogPath);
+        readStream.pipe(res);
+    } else {
+        const readStream = fs.createReadStream(compressedLogPath);
+        readStream.pipe(gzipStream).pipe(res);
+    }
+});
+
+router.get('/patches', function(req, res) {
```
ptc-mklaudiny, line 47: What kind of data are these patches? RGBD snapshots?

ptc-mklaudiny (approved):

routers/history.js
hobinjk-ptc: Yes, this is for persisting the snapshot along with the camera's pose on the server (in addition to the current persistence through window.localStorage)

libraries/recorder.js
hobinjk-ptc: Good point, thanks!

PR #701, opened 2023-02-20
Human tracking multiple cameras

ptc-mklaudiny:
Human tracking with fusion of poses from multiple cameras (devices running Toolbox app).

Related to PR in UI repo: https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/pull/391


libraries/HumanPoseFuser.js
```diff
@@ -0,0 +1,803 @@
...
+                        editorId: 0    // TODO: some server identificator
+                    };
+                    if (this.verbose) {
+                        console.log('assigning human obj=' + id + ' from ' + fusedObjectId);
+                    }
+                }
+            } else {
+                // whole group is unassigned, thus create new fused human object for it
+                // TODO?: add a string identifying the server  (equivalent to globalStates.tempUuid for devices)
+                let fusedObjectId = '_HUMAN_' + 'server' + '_pose1' + utilities.uuidTime();
```
ptc-mklaudiny, line 734: See TODO comment. Any suggestions about identificator of these server-based human objects? Do we actually need some server id unless we expect multiple servers in play for the same world object?

libraries/HumanPoseFuser.js
```diff
@@ -0,0 +1,803 @@
...
+        //let before = Object.entries(poseData).length;
+        let poseDataArr = Object.entries(poseData).filter(entry =>
+            (entry[1].joints.length > 0 && (this.objectsRef[entry[0]] !== undefined))
+        );
+        /*if (before != poseDataArr.length) {
+            console.log(`Pose data count: before=${before}, after=${poseDataArr.length}`);
+        }*/
+
+        // compare all pairs of poses and make binary matrix of spatial proximity
+        let proximityMatrix = Array.from(Array(poseDataArr.length), () => new Array(poseDataArr.length).fill(false));
```
hobinjk-ptc, line 589: nit: missing `new` on the first Array(poseDataArr.length) here

libraries/HumanPoseFuser.js
hobinjk-ptc: For the time being this feels reasonable enough. Existing unique things about the server are getIP and the edge agent network id, of which getIP is the easiest to use.

The main case I'm worried about (and can't test) is 2 phone servers both doing human pose fusing and both fused human pose objects showing up in the remote operator.

The most durable approach for the future would be to only fuse human pose objects if the world object they're positioned relative to is hosted on the current server. This approach could then use the world id as part of its id (and keep track of how many fused pose objects there are to differentiate pose1/pose2/etc.)

libraries/HumanPoseFuser.js
```diff
@@ -0,0 +1,803 @@
...
+                        };
+                    }
+                }
+
+                // setup active heartbeat for this new object
+                server.objectBeatSender(this.beatPort, fusedObjectId, this.ip, false, true);
+
+                // add new object to different global data structures
+                this.sceneGraphRef.addObjectAndChildren(fusedObjectId, this.objectsRef[fusedObjectId]);
+                utilities.writeObject(this.objectLookupRef, this.objectsRef[fusedObjectId].name, fusedObjectId);
```
hobinjk-ptc, line 759: It feels like there should be a better solution for creating a new object that's more of an all-in-one solution than all of these manual modifications and function calls but that can be left as future work for a more comprehensive server refactoring project.

hobinjk-ptc (approved):
LGTM, great work!

hobinjk-ptc:
Oh and this should be rebased/merged with master to see if the tests will pass

libraries/HumanPoseFuser.js
ptc-mklaudiny: I didn't know, that the system would currently allow to have the server on phone doing the human pose fusion. So the other phones would stream poses to some 'master' device.

The number in 'pose1' is currently redundant since we distinguish the human objects by 'hash' from creation time. Actually, it is bit odd that the object id building is split for the standard human objects made by the app. In UI [code](https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/5d6949ca0e4de41b3a308d8173ebd2648637f265/src/humanPose/index.js#L33), it creates '_HUMAN_device3zdi63z1_pose1'  and then the server [code](https://github.com/ptcrealitylab/vuforia-spatial-edge-server/blob/cab85be7478770a6bd388b5e43317dfd255840c2/server.js#L2490) expands it to '_HUMAN_device3zdi63z1_pose1_Lgk3swbv6nx'. Maybe we should move it all to UI on app side.


Once we start really supporting multiple people, then there will be several distinct poses at the same timestamp and we need the counter to distinguish them. This applies to the human objects created by the app and fused human objects created on server.

ptc-mklaudiny:
In terms of failing tests on Node 10.x, the problem is Object.fromEntries is supported from 12.x.  A broader question is how far back do we want to keep the support?

libraries/HumanPoseFuser.js
```diff
@@ -0,0 +1,803 @@
...
+                if (this.objectsRef[id] !== undefined) {
+                    this.objectsRef[id].parent = null;
+
+                    this.batchedUpdates[id] = {
+                        objectKey: id,
+                        frameKey: null,
+                        nodeKey: null,
+                        propertyPath: 'parent',
+                        newValue: null,
+                        editorId: 0    // TODO: some server identificator
```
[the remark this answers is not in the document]
benptc: Yeah that seems about right. We do have access to the utilities.uuidTime() function which will produce a random string, but we would need a modified version of the function to seed it with the server IP instead of the current timestamp, if we want it to return a consistent result per server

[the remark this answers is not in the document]
benptc: It should be technically correct with the current implementation, though, as the editorId is mainly used to prevent sending messages back to the toolbox app that originated the message. But since this message is coming from the server it doesn't need to be filtered this way.

libraries/HumanPoseFuser.js
```diff
@@ -0,0 +1,803 @@
...
+        let cutoffTS = latestTS - this.pastIntervalMs;
+
+        // delete poses older than a given ts across all human pose objects
+        for (let objPoses of Object.values(this.pastPoses)) {
+            this.removeOldPoseData(objPoses.poses, cutoffTS);
+        }
+    }
+
+    /**
+     * Removes poses of a given huamn object older than a specified timestamp.
```
benptc, line 326: nit: trailing space

libraries/HumanPoseFuser.js
benptc: @hobinjk-ptc This just made me realize – I'm not sure if the current apps are discovering other worlds that are hosted on the local iOS server of another phone. I've only been testing with multiple devices that connect to a server on an external laptop.

server.js
```diff
@@ -4098,6 +4141,10 @@ function deleteObject(objectKey) {
         utilities.deleteObject(objects[objectKey].name, objects, objectLookup, activeHeartbeats, knownObjects, sceneGraph, setAnchors);
     }
     // try to clean up any other state that might be remaining
+
+    if (objectKey.includes('_HUMAN_')) {
+        humanPoseFuser.removeHumanObject(objectKey);
+    }
```
benptc, line 4147: You might also need to add this logic to the `deleteObject` function in utilities.js (somehow we ended up with two deleteObject functions, which ideally would be consolidated into a single function)

benptc (approved):
Nice handling of complex logic. Great documentation. A few small comments left inline.

libraries/HumanPoseFuser.js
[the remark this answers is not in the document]
ptc-mklaudiny: Right. I did not encounter any issues with this so far. I will keep it as is.

PR #705, opened 2023-02-24
Improved speed of sending heartbeat

ptc-mklaudiny:
Improved speed of sending heartbeat of new objects from server to listeners. The objects appear quickly on client screens

### DIFF
--- a/server.js
+++ b/server.js
@@ -1177,17 +1177,9 @@ function serverBeatSender(udpPort, oneTimeOnly = true) {
  * @param {boolean} immediate if true the firdt beat will be sent immediately, not after beatInterval (works for one-time and periodic beats)
  **/
 
-function objectBeatSender(PORT, thisId, thisIp, oneTimeOnly, immediate) {
+function objectBeatSender(PORT, thisId, thisIp, oneTimeOnly = false, immediate = false) {
     if (isLightweightMobile) {
         return;
-    }
-
-    if (typeof oneTimeOnly === 'undefined') {
-        oneTimeOnly = false;
-    }
-
-    if (typeof immediate === 'undefined') {
-        immediate = false;
     }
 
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -1185,6 +1186,10 @@ function objectBeatSender(PORT, thisId, thisIp, oneTimeOnly) {
         oneTimeOnly = false;
     }
 
+    if (typeof immediate === 'undefined') {
+        immediate = false;
+    }
```
benptc, line 1191: I think you can give the parameter a default value in the function signature instead of handling it here

`function objectBeatSender(PORT, thisId, thisIp, oneTimeOnly, immediate = false) { ... `

server.js
```diff
@@ -1279,10 +1285,20 @@ function objectBeatSender(PORT, thisId, thisIp, oneTimeOnly) {
...
+        if (immediate) {
+            sendBeat();
+        }
+        // perturb the inverval a bit so that not all objects send the beat in the same time.
+        activeHeartbeats[thisId] = setInterval(sendBeat, beatInterval + utilities.randomIntInc(-250, 250));
+    } 
+    else {
         // Single-shot, one-time heartbeat
         // delay the signal with timeout so that not all objects send the beat in the same time.
+        let interval = immediate ? 0 : utilities.randomIntInc(1, 250);
```
benptc, line 1300: I might rename "interval" here to "delay", otherwise I expect it will repeat multiple times

benptc (approved):
Great to have this timing optimization! Left a couple comments inline but looks good.

server.js
ptc-mklaudiny: Thanks for the review. I defined default values for both oneTimeOnly and immediate, so it is more modern JS.

benptc: Thanks for bringing it up to date!

PR #714, opened 2023-03-01
Persist recorder log on exit instead of on sigint

hobinjk-ptc:
Running in the exit handler requires a synchronous persist function. PR also changes the behavior of the asynchronous version to never skip updates. Both this new approach and the old approach will lose data in the event of a compression/write error

### DIFF
--- a/libraries/recorder.js
+++ b/libraries/recorder.js
@@ -91,8 +91,7 @@ recorder.getCurrentLogName = function() {
     return 'objects_' + timeString + '.json';
 };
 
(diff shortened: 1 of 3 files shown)


ptc-mklaudiny (approved):
It makes sense to me. I tested it a bit with 30-40min recording where I shortened the interval for persisting to 10min. I don't know how it was performing before but there are small gaps in timestamps between the multiple saved json file. I saw 400 - 700ms gaps in data  (moreless confirmed also by lastUpdateDataTS which I store in human objects).

PR #716, opened 2023-03-02
Add API for manual log persistence

hobinjk-ptc:


dangond-ptc (commented):
Oops missed this but LGTM as long as it doesn't console.log too much

PR #717, opened 2023-03-03
adds new DELETE /object/${objectID} route for manager to use

benptc:
Deprecates the old way to delete objects, which was with:

POST /
{
  action: 'delete',
  name: `${object.name}`,
  frame: ''
}

You can now delete an object with:

DELETE `/object/${objectID}`

This is consistent with how to delete frames (tools), which is currently done with:

DELETE `/object/${objectID}/frames/${frameID}`

### DIFF
--- a/controllers/object.js
+++ b/controllers/object.js
@@ -12,6 +12,35 @@ var objectsPath;
 var identityFolderName;
 var git;
 var sceneGraph;
+// needed for deleteObject
+let objectLookup;
+let activeHeartbeats;
+let knownObjects;
+let setAnchors;
+
+const deleteObject = function(objectID) {
+    let object = utilities.getObject(objects, objectID);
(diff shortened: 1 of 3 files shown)


controllers/object.js
```diff
@@ -12,6 +12,36 @@ var objectsPath;
...
+let knownObjects;
+let setAnchors;
+
+const deleteObject = function(objectID) {
+    let object = utilities.getObject(objects, objectID);
+    if (!object) {
+        return {
+            status: 404,
+            error: `Object ${objectID} not found`
+        }
```
hobinjk-ptc, line 27: see lint warnings inline

hobinjk-ptc (approved):

ktheo-ptc (approved):
Looks good to me!

PR #718, opened 2023-03-03
delete avatar object if we haven't heard its heartbeat in the last 5 seconds

benptc:
Avatar heartbeats come every 1 second with https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/pull/367, so deleting an avatar if no heartbeats within the past 5 seconds seems reasonable to me.


hobinjk-ptc (approved):
wcgw

benptc:
Haha, thanks for the review. We'll hope for the best

PR #721, opened 2023-03-06
increase avatar deletion age back up to 15s; 5s led to some accidental deletions

benptc:
If the app hangs for too long your own avatar could be deleted by your server, causing your cursor and presence to disappear. No mechanism to re-establish it yet if it gets deleted, so I'm giving it a more generous time buffer to minimize these instances for now.


dangond-ptc (approved):
LGTM!

PR #732, opened 2023-03-14
Web gl multi frame loading

ptc-rdeleeuw:
implementing command buffers, state tracking and context lost/restore for the webgl implementation 

should be merged together with:
https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/pull/379
https://github.com/ptcrealitylab/vuforia-spatial-core-addon/pull/218
https://github.com/ptcrealitylab/vuforia-spatial-remote-operator-addon/pull/95


libraries/objectDefaultFiles/glCommandBuffer.js
```diff
@@ -0,0 +1,1073 @@
...
+         * @type {CommandBuffer[]}
+         */
+        this.commandBuffers = [];
+
+        /**
+         * the commandbuffer currently used for repeated rendering
+         * the rendering buffer starts as an empty commandbuffer, so the tool doesn't show until it is fully loaded.
+         * @type {CommandBuffer}
+         */
+        this.renderCommandBuffer = new CommandBuffer(workerId, -1, true, []);
```
hobinjk-ptc, line 1010: What is the effect of initializing a commandbuffer with synclock set to -1 here? It looks like this is a placeholder so `null` would be a bit clearer in purpose. It also looks like CommandBuffer doesn't handle an optional synclock

vuforia-spatial-edge-server.code-workspace
```diff
@@ -0,0 +1,10 @@
+{
```
hobinjk-ptc, line 1: this and the njsproj should be added to gitignore and not included

hobinjk-ptc (commented):
Excellent and comprehensive work! Some comments inline

libraries/objectDefaultFiles/glCommandBuffer.js
```diff
@@ -0,0 +1,1073 @@
...
+                    let buffer = new SharedArrayBuffer(buffer_length);
+                    this.addMessageAndWait("getActiveAttrib", [program, index], buffer);
+                    // decode result
+                    let utf8TextDecoder = new TextDecoder();
+                    let texbuf = new Uint8Array(buffer_length - 8);
+                    texbuf.set(new Uint8Array(buffer, 0, buffer_length - 8));
+                    let sizeTypeBufMem = new ArrayBuffer(8);
+                    new Uint8Array(sizeTypeBufMem).set(new Uint8Array(buffer, buffer_length - 8, 8));
+                    let sizeTypeBuf = new Int32Array(sizeTypeBufMem);
+                    return {name: utf8TextDecoder.decode(texbuf).substring(0, buffer_length - 9), size: sizeTypeBuf[0], type: sizeTypeBuf[1]}
```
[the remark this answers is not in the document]
ptc-rdeleeuw: Let me add some more documentation like I did there. 

it's the size needed to receive the structure:
https://developer.mozilla.org/en-US/docs/Web/API/WebGLActiveInfo

libraries/objectDefaultFiles/glCommandBuffer.js
```diff
@@ -0,0 +1,1073 @@
...
+            } else {
+                // if an unknown function is called, execute it as a default command, not expecting any type of response
+                this.gl[fnName] = (...args) => {
+                    this.addMessage(fnName, Array.from(args));
+                }
+            }
+        }
+        // copy over the gl constants to our fake webgl implmentation
+        for (const constName in message.constants) {        
+            if (this.gl.hasOwnProperty(constName)) {
```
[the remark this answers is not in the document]
ptc-rdeleeuw: strange this code didn't change that much from the original bootstrap code, but you're correct. Let me correct this.

libraries/objectDefaultFiles/glCommandBuffer.js
ptc-rdeleeuw: I don't know why I made it -1, null indeed is much better, this code is server side only and makes sure there is always an empty render buffer to render from. This buffer doesn't need sync message support because it is never send.

PR #741, opened 2023-03-27
Make server work on non-UDP-compatible networks

benptc:
If the UDP client.send fails for a heartbeat or action message, tries to send the message to connected clients over an existing websocket.


dangond-ptc (approved):
Looks good to me!

dangond-ptc:
Wait no, there's an error in there :|

dangond-ptc (changes requested):

benptc:
> Wait no, there's an error in there :|

Just realized that too... will fix soon

PR #748, opened 2023-04-03
Reliably sync envelope state

benptc:
Mainly updates the:
1. Don't overwrite an existing node if `/addNode/` is called more than once
2. When calling `/addNode/`, allow the user to set default values for any node properties, such as data.value
3. Trigger `socketHandler.sendDataToAllSubscribers` when a node is added, to ensure that iframes loaded in any order will eventually get the node's default value sent to them, even if they loaded before the node was added.
4. Adds a new parameter to Envelope constructor, `opensWhenAdded`, that will default the "open" node to 1 instead of 0 when the tool is first created. (Note: planning to refactor this constructor to take an `options` object instead, but I'm limiting the scope of this PR so I don't cause backwards-compatibility issues or large number of lines changed)
5. Prevent recursive `realityInterface.write` calls that get made when an envelope is opened/closed in response to reading an open/close message from the "open" node.
6. Refactor the redundant `initNode` and `sendCreateNode` realityInterface APIs to trigger the same message in the userinterface.

Corresponding updates coming soon today for userinterface and addons...


libraries/objectDefaultFiles/envelope.js
```diff
@@ -559,13 +567,13 @@
             if (typeof this.lastOpenValue === 'undefined') {
                 this.lastOpenValue = event.value;
             }
-            if (this.lastOpenValue === event.value) {
+            if (this.lastOpenValue === 0 && event.value === 0) {
                 return; // prevents it from closing itself when the node first loads or on duplicate data
             }
             if (event.value < 0.5) {
-                this.close();
+                this.close({ dontWrite: true });
```
hobinjk-ptc, line 574: Good cases to eliminate!

hobinjk-ptc (approved):
LGTM, excellent to clean up all these behaviors!

PR #751, opened 2023-04-04
Replace use of Array.prototype.at(-1) with equivalent indexing

hobinjk-ptc:
Node versions less than 16 don't support this function


benptc (approved):

ptc-mklaudiny (commented):
Thanks for catching it. I suppose that the CI tests passed because basic run of server didn't go through this code path.  
It might be good to have some kind of static code analysis to catch compatibility of JS features with different Node version.

PR #763, opened 2023-04-12
Always uses direct triggerUDPCallbacks logic; Only calls triggerUDPCallbacks from message handler if a remote message

hobinjk-ptc:
triggerUDPCallbacks from send is lower latency than waiting for a roundtrip and there are some network conditions where a send will successfully complete but not bounce back for the listener to receive.


benptc (approved):
Seems reasonable, LGTM

PR #771, opened 2023-04-20
only utilizes the open node of an envelope if the tool isnt running in full2d mode

benptc:
This seems to be the least invasive way to make the envelopes with a 3D scene sync across clients, while making the fully-2D envelopes (chat, search, etc) personal to each user.

Goes with https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/pull/421

### DIFF
--- a/libraries/objectDefaultFiles/envelope.js
+++ b/libraries/objectDefaultFiles/envelope.js
@@ -169,7 +169,9 @@
             compatibleFrameTypes: this.compatibleFrameTypes
         });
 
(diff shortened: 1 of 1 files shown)


hobinjk-ptc (approved):
Oh interesting, this is one way of doing it. Makes sense, LGTM!

benptc:
> Oh interesting, this is one way of doing it. Makes sense, LGTM!

Yeah I wasn't originally planning to solve it this way, but this seems to work well enough without overcomplicating things.

PR #772, opened 2023-04-20
Human tracking updated by children property

ptc-mklaudiny:
Added new property 'updatedByChildren' to only fused human objects. This represents a current best human object updating the fused object. It is designed as array because it can be tru fusion of multiple objects (views) in next iteration.
Possible states of the property are: 
undefined  - for standard human objects generated by toolbox apps
['none'] - for fused human objects; no object currently updating fused object  (unlikely but possible scenario)
['\_HUMAN\_**deviceVJ4ubfsh**_pose1__O26kv6p852m']  - for fused human objects; objectId of standard human object which is currently selected.

The bold substring (from editor uuid) should be used to derive which camera/app is currently driving the fused human object. For the standard objects which don't have the property defined , this can be extracted from their objectId. This can be the basis for coloring skeletons in the new lens showing currently active camera.

In Acceleration lens, any change of this property from one pose to another in consecutive timestamps indicates possible artificial 'jump'. So we could choose not to compute/show any acceleration at that point of time.

### DIFF
--- a/libraries/HumanPoseFuser.js
+++ b/libraries/HumanPoseFuser.js
@@ -548,13 +548,18 @@ class HumanPoseFuser {
 
         // detect change in the subset of associated (child) human objects which currently update their fused object
         const equalArrays = (a, b) => a.length === b.length && a.every((element, index) => element === b[index]);
+
+        // because of recorder limitations this cannot be an empty array. Instead we place 'none' item.
         let arr = [];
         if (selectedObjectId) {
             arr.push(selectedObjectId);
         }
+        else {
+            arr.push('none');
+        }
         if (!equalArrays(arr, this.objectsRef[fusedObjectId].updatedByChildren)) {
             // a change detected, we need to update the property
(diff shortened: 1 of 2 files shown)


libraries/HumanPoseFuser.js
```diff
@@ -140,12 +140,14 @@ class HumanPoseFuser {
...
-        /** Dictionary which collects all updates to 'parent' property of human objects in current run of fusion. It is a dictionary so we keep just one change per human object */
+        /** Dictionary which collects all updates to properties of human objects in current run of fusion (currently properties 'parent' and 'updatedByChildren')
+         * It is a dictionary of dictionaries so we keep just one change per human object per its property
+         */
         this.batchedUpdates = {};
 
         /* Configuration parameters */
         /** Verbose logging */
-        this.verbose = false;
+        this.verbose = true;
```
dangond-ptc, line 150: This should probably be set back to false before merging.

PR #790, opened 2023-05-05
Human tracking multiview triangulation

ptc-mklaudiny:
I introduced new methods of fusing human objects from multiple views. More testing and tuning is needed to select the most reliable one in near term (eg. LiveWorx). It's easy to switch between them through flags without changing the current behaviour. I would like to have this merged since lot of the functionality will be useful down the line. 

Also, added new dependency  ml-matrix for linear algebra.

### DIFF
--- a/libraries/HumanPoseFuser.js
+++ b/libraries/HumanPoseFuser.js
@@ -3,6 +3,7 @@ const HumanPoseObject = require('./HumanPoseObject');
 const sgUtils = require('./sceneGraph/utils.js');
 const utilities = require('./utilities.js');
 const server = require('../server');
+//const { Matrix } = require('ml-matrix');
 
 /** Joint schema of human pose used for creation of fused HumanPoseObjects. This schema is also expected from the human objects coming from UI code of  ToolboxApp. */
 const JOINTS = {
@@ -30,12 +31,21 @@ const JOINTS = {
     PELVIS: 'pelvis', // synthetic
 };
 
+const FusionMethod = Object.freeze({
+    BestSingleView: 'BestSingleView',
+    MultiViewTriangulation: 'MultiViewTriangulation'
+});
+
 /**
  * The object with pose data incoming from Toolboxapps
  * @typedef {Object} WholePoseData
  * @property {string} name
- * @property {number} timestamp
  * @property {Array.<{x: number, y: number, z: number, confidence: number}> } joints
+ * @property {number} timestamp
+ * @property {Array.<number>} imageSize
+ * @property {Array.<number>} focalLength
+ * @property {Array.<number>} principalPoint
+ * @property {Array.<number>} transformW2C
  */
 
 /**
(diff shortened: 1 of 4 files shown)


libraries/HumanPoseFuser.js
```diff
@@ -403,38 +550,51 @@ class HumanPoseFuser {
...
+     * @param {number} distance - distance between two poses (as computed by calculateDistanceBetweenPoses)
+     * @param {number} ts1 - timestamp of first pose
+     * @param {number} ts2 - timestamp of second pose
+     * @param {number} multiplier - multiplier of distance threshold
+     * @return {boolean}
+     */
+    arePosesOfSamePerson(distance, ts1, ts2, multiplier = 1.0) {
+        // distance threshold is calculated from two components:
+        // - base deviation between poses at the same timestamp
+        // - max possible shift between poses from different timestamp due to person's movement
```
hobinjk-ptc, line 563: nit: trailing spaces (see Action output below)

libraries/HumanPoseFuser.js
```diff
@@ -526,16 +686,242 @@ class HumanPoseFuser {
...
+    triangulatePose(poseDataArr) {
+        if (poseDataArr.length < 2) {
+            return null;
+        }
+
+        // prepare data
+        let projectionMatrices = []; // per-view
+        let jointsInView = []; // joint 2D positions + confidences per view
+        let latestTS = 0; // the latest data timestamp
+        for (let [id, pose] of poseDataArr) {
```
hobinjk-ptc, line 767: another nit inline, could be _id to make it clear that you don't need the id from poseDataArr's entries

libraries/HumanPoseFuser.js
```diff
@@ -568,33 +1015,29 @@ class HumanPoseFuser {
...
-        this.objectsRef[fusedObjectId].updateJoints(poseData[selectedObjectId].joints);
-        this.objectsRef[fusedObjectId].lastUpdateDataTS = poseData[selectedObjectId].timestamp;
+        this.objectsRef[fusedObjectId].updateJoints(finalPose.joints);
+        this.objectsRef[fusedObjectId].lastUpdateDataTS = finalPose.timestamp;
         server.resetObjectTimeout(fusedObjectId); // keep the object alive
 
-        // create copy and update name (confidence attribute not copied)
-        let wholePose = Object.assign({}, poseData[selectedObjectId]);
+        // create copy and update name (unnecessary properties not copied)
+        let wholePose = Object.assign({}, finalPose);
```
hobinjk-ptc, line 1035: I don't see any bugs resulting from this but note that Object.assign is a shallow copy, for example mutating finalPose.joints will change wholePose.joints and vice versa

hobinjk-ptc (approved):
Good work! Tested this out locally with 2 devices and LGTM

libraries/HumanPoseFuser.js
ptc-mklaudiny: It should be fine, the pose data like joints are not expected to change one generated. In some cases, internal 'joints' array might be pointed at from several WholePoseData objects with different lifetimes. But I assume that garbage collector does a right thing and removes it once there is no reference to it.

PR #802, opened 2023-05-11
Trigger reloadObject if a remote client sends a property update of an unknown object

hobinjk-ptc:
This should make the client realize its object got deleted and it should refresh its local state.

### DIFF
--- a/server.js
+++ b/server.js
@@ -3503,6 +3503,13 @@ function socketServer() {
         socketHandler.socket = socket;
         //console.log('connected to socket ' + socket.id);
 
+        /**
+         * @type {{[objectKey: string]: bool}}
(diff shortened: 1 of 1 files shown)


benptc (approved):
Good idea, LGTM

ptc-mklaudiny (approved):
Nice addition.
It could be extended to public data channel. When there is a disynchronisation of objects between app and server for human pose objects, the pose updates get rejected by the following condition.
https://github.com/ptcrealitylab/vuforia-spatial-edge-server/blob/b2e4909effd8f29b69f742078e3a696acf44c9d1/server.js#L3721

Steve-KX-RL (approved):
LGTM!

PR #803, opened 2023-05-12
Added new mode for human object fusion 

ptc-mklaudiny:
Best single-view poses are translated to be close to a pose linearly interpolated from pose from multiple views.

This is a backup method behind a flag to have more alternatives if we see too big jumps when the best camera is switched at Liveworx.

### DIFF
--- a/libraries/HumanPoseFuser.js
+++ b/libraries/HumanPoseFuser.js
@@ -34,6 +34,7 @@ const JOINTS = {
 /** Enum for the method for fusing poses for the same person from several views/toolbox apps. */
 const FusionMethod = Object.freeze({
     BestSingleView: 'BestSingleView',
+    BestSingleViewWithCorrection: 'BestSingleViewWithCorrection',
     BestSingleViewWithMultiviewCorrection: 'BestSingleViewWithMultiviewCorrection',
     MultiViewTriangulation: 'MultiViewTriangulation'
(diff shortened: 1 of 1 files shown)


libraries/HumanPoseFuser.js
```diff
@@ -189,7 +190,7 @@ class HumanPoseFuser {
         /** max velocity of whole body - picked 2 m/s (average walking speed is 1.4 m/s) */
         this.maxHumanVelocity = 2.0;  // unit: mm/ms
         /** method for fusing poses for the same person from several views/toolbox apps. */
-        this.fusionMethod = 'BestSingleView'; // 'BestSingleViewWithMultiviewCorrection';
+        this.fusionMethod = 'BestSingleView'; // 'BestSingleViewWithCorrection'; // 'BestSingleViewWithMultiviewCorrection';
```
hobinjk-ptc, line 193: Since you have FusionMethod available above, it would be more typo-redundant (and, depending on ide support, enable autocomplete) to write this as FusionMethod.BestSingleView instead of the string constant

hobinjk-ptc (approved):

libraries/HumanPoseFuser.js
ptc-mklaudiny: Right, good point.

PR #850, opened 2023-07-17
Oauth support

dangond-ptc:


dangond-ptc:
https://github.com/ptcrealitylab/vuforia-spatial-toolbox-ios-swift/pull/439
https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/pull/495
https://github.com/ptcrealitylab/pop-up-onboarding-addon/pull/75

libraries/serverHelpers/oauthRequestHandlers.js
```diff
@@ -0,0 +1,54 @@
+const fetch = require('node-fetch');
+const querystring = require('querystring');
+
+const oauthRefreshRequestHandler = (req, res) => {
+    const refreshUrl = req.params[0]; // TODO: get this from the tool somehow to prevent leaking secret to any supplied url
```
hobinjk-ptc, line 5: If this is expected to always be an onshape url there could be an allowList of known urls to which we're okay with sending oauth requests. Any new developer of an oauth-enabled tool would then locally modify this allowList and eventually get us to ad the new domain to all edge servers

hobinjk-ptc (approved):

libraries/serverHelpers/oauthRequestHandlers.js
dangond-ptc: I was thinking of having the allowList be part of the secrets file for the add-on, since this is designed to be used with anything that supports OAuth, not just Onshape.

hobinjk-ptc: Thumbs up emoji

benptc (approved):
LGTM

PR #858, opened 2023-07-28
Cleaned up logging

dangond-ptc:


libraries/utilities.js
```diff
@@ -539,25 +504,13 @@ exports.generateChecksums = function generateChecksums(objects, fileArray) {
 
 
 exports.updateObject = function updateObject(objectName, objects) {
-    console.log('update ', objectName);
-
     var objectFolderList = fs.readdirSync(objectsPath).filter(function (file) {
-        return fs.statSync(path.join(objectsPath, file)).isDirectory();
+        return fs.statSync(path.join(objectsPath, file)).isDirectory() && file[0] !== '.';
```
benptc, line 508: Nice little cleanup thrown in here 👍

libraries/utilities.js
```diff
@@ -625,11 +574,9 @@ exports.updateObject = function updateObject(objectName, objects) {
 };
 
 exports.deleteObject = function deleteObject(objectName, objects, objectLookup, activeHeartbeats, knownObjects, sceneGraph, setAnchors) {
-    console.log('Deleting object: ' + objectName);
-
     let objectFolderPath = path.join(objectsPath, objectName);
     if (fs.existsSync(objectFolderPath)) {
-        fs.rmdirSync(objectFolderPath, {recursive: true});
+        fs.rmSync(objectFolderPath, {recursive: true});
```
benptc, line 579: I thought fs.rmSync wasn't added to node until v14.x, so this would cause the 12.x build to fail, but it doesn't seem to be failing so maybe it's ok?

benptc (approved):
Thank you for the thorough cleanup job!

libraries/utilities.js
dangond-ptc: Oh, that's a good catch actually. I changed it due to the deprecation warnings, but if we're still supporting 12.x, I can revert this. The only reason it's not failing the build is that the server doesn't delete anything on startup, and I think the test only checks to see if the server can run for 10 seconds without crashing.

hobinjk-ptc: The iPhones run 12 (they can be updated to 16 but that has not yet happened)

dangond-ptc: OK, will revert for now

PR #862, opened 2023-08-07
Add area target and main camera listeners

Steve-KX-RL:


libraries/objectDefaultFiles/object.js
```diff
@@ -728,6 +728,9 @@
                 this.promptForArea = makeSendStub('promptForArea');
                 this.getEnvironmentVariables = makeSendStub('getEnvironmentVariables');
                 this.getUserDetails = makeSendStub('getUserDetails');
+                this.getAreaTargetMesh = makeSendStub('getAreaTargetMesh');
```
dangond-ptc, line 731: Did you use this method in the measurement tool? I wasn't able to find it. I'm curious as to how it performs with larger world objects.

Steve-KX-RL: Not yet. I was planning to use it for a future volumetric mode (compute the area target mesh volume within a polygonal volume).

dangond-ptc (approved):

benptc (approved):
LGTM. Good work with the new APIs!

PR #875, opened 2023-08-24
Add statusCode to local proxy responses

hobinjk-ptc:
Resolves an issue where chrome would interpret the (incorrect) 200 status code on a partial (206) response as meaning the partial response was a complete response. This correctly forwards the 206 status code in the proxy, fixing the issue with Chrome.


dangond-ptc (approved):
Good catch! Can't believe I missed that!

PR #881, opened 2023-08-30
Update object.js

Steve-KX-RL:
Added various listeners (for spatial cursor shader, AreaTarget shader, and height label) when measurement tool state changes.


libraries/objectDefaultFiles/object.js
```diff
@@ -2151,6 +2158,36 @@
...
+            postDataToParent({
+                turnMeasureMapUI: boolean
+            });
+        }
+
+        this.toggleMeasureMapUI = function() {
+            postDataToParent({
+                toggleMeasureMapUI: true
+            });
+        }
```
benptc, line 2190: Can you add a comment to each of these APIs documenting what they do? It will be useful in future to not forget what these do.

libraries/objectDefaultFiles/object.js
```diff
@@ -2151,6 +2158,36 @@
...
+
+        this.toggleCrossRotation = function(boolean) {
+            postDataToParent({
+                toggleCrossRotation: boolean
+            });
+        }
+        
+        this.toggleCloseLoop = function(boolean) {
```
benptc, line 2174: I'm guessing this API relates to changing the appearance of the spatial cursor, but that's only because I know what you've been working on – can we rename all the APIs that affect the spatial cursor appearance to be a bit more clear that they're affecting the cursor appearance?

libraries/objectDefaultFiles/object.js
```diff
@@ -509,6 +509,7 @@
 
                             postDataToParent({
                                 pointerDownResult: 'interaction'
+                                // refer to this bit of code
```
benptc, line 512: This comment looks like it can be removed

benptc (changes requested):
Overall good, but a few comments about documentation

dangond-ptc (changes requested):
Looks good, same comments as Ben

dangond-ptc (approved):

libraries/objectDefaultFiles/object.js
```diff
@@ -2280,8 +2293,8 @@
                     name: processTitle,
                     numStopsRequired: options.numStopsRequired // if included, stop will need to be called this many times
                 }
-            });
-        };
+            })
+        }
```
benptc, line 2297: nit: missing semicolons here and throughout the file in the new functions you've added (you can check the code on github or do `eslint .` locally (if you've installed eslint) to see where the linting system is complaining)

benptc (approved):
Thanks for adding the documentation to these APIs!

PR #882, opened 2023-08-31
adds objectDefaultFiles/styles directory to contain any shared css files

benptc:
This is a very quick first proposal for where to put shared styles that multiple tools can make use of. Open to suggestions.

Adds a common.css file that for now just contains the tool-color-gradient style for the minimized icon shimmer

Tools can use these files like so: `<link rel="stylesheet" href="objectDefaultFiles/styles/common.css">`
(example: https://github.com/ptcrealitylab/pop-up-onboarding-addon/pull/84)


hobinjk-ptc (approved):

ktheo-ptc (approved):
Great to have this!

PR #905, opened 2023-09-28
Fix all eslint warnings

hobinjk-ptc:
All changes are expected to not change the behavior of the code, only make it look nicer


benptc (approved):
Looks good! Tested locally too

PR #909, opened 2023-10-03
Set local ip on ios

hobinjk-ptc:
Requires https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/pull/528


dangond-ptc (approved):

hobinjk-ptc:
I said fixes instead of a normal non-github word 🤦

PR #913, opened 2023-10-06
spatialInterface.captureSpatialSnapshot API

benptc:
Adds new API for tools to programmatically add spatial snapshots to the scene. Current userinterface implementation only works on AR devices, not on remote operator, but that's something I'll explore more in the userinterface before making that corresponding PR.

Goes with https://github.com/ptcrealitylab/vuforia-spatial-toolbox-ios-swift/pull/495


libraries/objectDefaultFiles/object.js
```diff
@@ -1825,6 +1826,28 @@
...
+         */
+        this.captureSpatialSnapshot = function() {
+            postDataToParent({
+                captureSpatialSnapshot: true
+            });
+            return new Promise((resolve, reject) => {
+                spatialObject.messageCallBacks.captureSpatialSnapshotResult = function (msgContent) {
+                    if (typeof msgContent.spatialSnapshotData !== 'undefined') {
+                        resolve(msgContent.spatialSnapshotData);
+                        delete spatialObject.messageCallBacks['captureSpatialSnapshotResult']; // only trigger it once
```
hobinjk-ptc, line 1841: `delete` can take `spatialObject.messageCallBacks.captureSpatialSnapshotResult` without quotes which may help some static analysis

hobinjk-ptc (approved):

PR #919, opened 2023-10-10
Constant refactoring

hobinjk-ptc:
Refactors out constants to new constants.js file. Throws in a small refactor to share more code in Addons and a test thereof.

yes I am very proud of this pun

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -1,5 +1,7 @@
-exports.identityFolderName = '.identity';
-// The protocol of this server
-exports.protocol = 'R2';
-// The version of this server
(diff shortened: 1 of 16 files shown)


constants.js
dangond-ptc, line 1: I realize now that I missed this earlier, but we should consider doing the following to ensure the constants stay constant:
```
module.exports = Object.freeze({
  // Stuff we want to export as constants
})
```

dangond-ptc (approved):

constants.js
hobinjk-ptc: Done, thanks!

PR #920, opened 2023-10-10
Fix crash due to race condition during target zip upload

hobinjk-ptc:
Under some conditions the jpg contained in the target zip has already been moved/processed before the call to `rename`


dangond-ptc (commented):
Nice catch!

PR #921, opened 2023-10-10
 Add test for target uploading 

hobinjk-ptc:
There was a little scooch down into madness getting the test to work as an action but it's good* now.

*passed exactly once

### DIFF
--- a/server.js
+++ b/server.js
@@ -2942,7 +2942,11 @@ function objectWebServer() {
                                                 let deferred = false;
                                                 function finishFn(folderName) {
                                                     return async function() {
-                                                        await fsProm.rmdir(path.join(folderD, identityFolderName, 'target', folderName));
+                                                        try {
+                                                            await fsProm.rmdir(path.join(folderD, identityFolderName, 'target', folderName));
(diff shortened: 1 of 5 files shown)


server.js
```diff
@@ -2942,7 +2942,11 @@ function objectWebServer() {
                                                 let deferred = false;
                                                 function finishFn(folderName) {
                                                     return async function() {
-                                                        await fsProm.rmdir(path.join(folderD, identityFolderName, 'target', folderName));
+                                                        try {
+                                                            await fsProm.rmdir(path.join(folderD, identityFolderName, 'target', folderName));
```
dangond-ptc, line 2946: Why is `{recursive: true}` needed sometimes and not others?

dangond-ptc (approved):

server.js
hobinjk-ptc: Here it's because the folder is expected to be empty (with all of its contents processed by the renaming functions) and it would be evidence of an error condition if it weren't. This is definitely a special boy, most other uses of rmdir could get `recursive: true` and not explode

PR #923, opened 2023-10-13
Fs prom exists fix

dangond-ptc:
Server crashed when loading analytics data from history due to fsProm.exists not...existing. Replaced it with fsProm.stat, which is the recommended way of checking if a file exists. Also ensured createReadStream continues to use old-fashioned `fs`, since fsProm.createReadStream also does not exist.

### DIFF
--- a/routers/history.js
+++ b/routers/history.js
@@ -6,13 +6,14 @@ const stream = require('stream');
 const zlib = require('zlib');
 
 const recorder = require('../libraries/recorder.js');
+const {fileExists} = require('../libraries/utilities.js');
 
 const router = express.Router();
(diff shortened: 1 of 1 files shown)


hobinjk-ptc:
As you saw in your commit, `stat` throws an error on nonexistence which is a bit annoying. I have the `fileExists` function defined [here](https://github.com/ptcrealitylab/vuforia-spatial-edge-server/blob/master/libraries/utilities.js#L1082) which wraps this.

hobinjk-ptc (approved):
Thanks for catching these! Left a comment with a recommended change

PR #924, opened 2023-10-13
Renamed hydration method to better represent new functionality of sen…

dangond-ptc:
…ding all data, not just region cards

### DIFF
--- a/libraries/objectDefaultFiles/object.js
+++ b/libraries/objectDefaultFiles/object.js
@@ -1753,7 +1753,7 @@
         };
 
         /**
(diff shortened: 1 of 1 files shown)


benptc (approved):

hobinjk-ptc (approved):

libraries/objectDefaultFiles/object.js
```diff
@@ -1755,11 +1755,11 @@
         /**
          * @param {Array<{startTime: number, endTime: number}>} regionCards
```
hobinjk-ptc, line 1756: out of date comment

PR #933, opened 2023-10-24
Filter out all interfaces on mobile that aren't en0

hobinjk-ptc:


benptc (approved):
Tested locally on cellular and wifi – both were able to scan/upload and successfully share the metaverse with a remote operator

PR #939, opened 2023-11-07
notification style added

ktheo-ptc:
- notification styles added to common.css


benptc (approved):
Nice to have this here now!

PR #941, opened 2023-11-09
Prevent cycles from setParent; Fix overwriting worldId of world object

hobinjk-ptc:
Fixes an issue where setParent would introduce a cycle into the scene graph due to treating secondary world objects like children of the first observed world object. Honestly not sure what the correct behavior would be, but I'm at least confident that it's not get stuck recursing forever.


benptc (approved):
Nice safety guards ... I've definitely occasionally had my world's worldId set to something incorrect before

PR #947, opened 2023-11-16
Safer logic around fs access and async/await

hobinjk-ptc:
Various calls can throw exceptions if e.g. trying to delete a file which
already doesn't exist, so this wraps all patterns where condition is
checked -> condition is acted on with exception handlers for the case
where the condition changes during function execution. Introduces
mkdirIfNotExists and unlinkIfExists for this purpose

Tries to identify places where changes in a global variable (`objects`,
it's always `objects`) could happen during the execution of an async
function and break preconditions (usually the deletion of objects[key]).
Also wraps some potentially problematic calls to fs apis that could
throw.

### DIFF
--- a/server.js
+++ b/server.js
@@ -343,7 +343,7 @@ if (!isLightweightMobile) {
 
 // This file hosts all kinds of utilities programmed for the server
 const utilities = require('./libraries/utilities');
-const {fileExists} = utilities;
+const {fileExists, mkdirIfNotExists, unlinkIfExists} = utilities;
 const nodeUtilities = require('./libraries/nodeUtilities');
 const recorder = require('./libraries/recorder');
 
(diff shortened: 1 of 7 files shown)


server.js
```diff
@@ -1232,16 +1259,18 @@ async function objectBeatSender(PORT, thisId, thisIp, oneTimeOnly = false, immed
...
         let tdtPath = path.join(targetDir, 'target.3dt');
         var fileList = [jpgPath, xmlPath, datPath, glbPath, tdtPath];
-        objects[thisId].tcs = await utilities.generateChecksums(objects, fileList);
-    }
-
-    // if no target files exist, checksum will be undefined, so mark with checksum 0 (anchors have this)
-    if (typeof objects[thisId].tcs === 'undefined') {
-        objects[thisId].tcs = 0;
+        const tcs = await utilities.generateChecksums(objects, fileList);
+        if (objects[thisId]) {
```
ptc-mklaudiny, line 1263: Question for my better understanding of underlying execution model of JS. 
These extra checks whether an object exists (so there is a valid access to it later in the function) are necessary only when there is asynchronous call beforehand in the currently executed function. Although, the current execution waits due to await, there is a switch of context by JS engine to execute another function. There the object can be modified/deleted and once the await finishes, there is a problem in the current function.

When there is no async call in the function, there is just one check necessary. The context switch by JS engine can happen only after the current function is finished.

ptc-mklaudiny (approved):

ptc-mklaudiny:
Thanks for robustifying the server code. I will try it out in my tests which triggered this effort.

server.js
hobinjk-ptc: Yes, that's correct

PR #958, opened 2023-12-11
Move remote operator signalling server into edge server

hobinjk-ptc:


server.js
```diff
@@ -449,12 +452,16 @@ function Protocols() {
...
+                    console.warn('protocol: Unable to find objectData', message);
+                }
+
                 return {
                     object: msgContent.object,
                     frame: msgContent.frame,
                     node: msgContent.node,
                     logic: msgContent.logic,
-                    data: this.objectData.data
+                    data: this.objectData.data || msgContent.data,
```
benptc, line 464: Since this doesn't return early if !this.objectData, should it be something like `(this.objectData && this.objectData.data) || msgContent.data` (similar to what you did in messageToSend)?

benptc (commented):
Signalling server looks good - nice and elegant implementation 🥇

server.js
hobinjk-ptc: Good point! Updated accordingly

benptc (approved):

PR #967, opened 2024-01-04
Adds targetId to object.json; loosens restriction that ID from target.xml must match the objectID

benptc:
`targetId` is extracted from the target.dat file by unzipping target.dat and parsing the contained config.info file.

Goes with https://github.com/ptcrealitylab/vuforia-spatial-toolbox-ios-swift/pull/555

### DIFF
--- a/libraries/utilities.js
+++ b/libraries/utilities.js
@@ -234,7 +234,6 @@ exports.getAnchorIdFromObjectFile = getAnchorIdFromObjectFile;
 
 exports.getTargetIdFromTargetDat = async function getTargetIdFromTargetDat(targetFolderPath) {
     return new Promise((resolve, reject) => {
-        // deferred = true;
         // unzip the .dat file and read the unique targetId from the config.info file
         let unzipperDat = new DecompressZip(path.join(targetFolderPath, 'target.dat'));
 
@@ -244,14 +243,10 @@ exports.getTargetIdFromTargetDat = async function getTargetIdFromTargetDat(targe
         });
 
         unzipperDat.on('extract', async function () {
-            // finish();
-            console.log('.dat Unzipper finished');
-
             let configFilePath = path.join(targetFolderPath, 'config.info');
             if (await fileExists(configFilePath)) {
-                // try to read the config.info file as text
-                let targetUniqueId = await getTargetIdFromFile(configFilePath);
-                console.log('targetUniqueId = ', targetUniqueId);
+                // read the id stored within the config.info file (it's actually structured as XML)
+                let targetUniqueId = await getTargetIdFromConfigFile(configFilePath);
                 console.log('TODO: cleanup config.info file instead of leaving it in the folder');
                 resolve(targetUniqueId);
             } else {
@@ -275,7 +270,7 @@ exports.getTargetIdFromTargetDat = async function getTargetIdFromTargetDat(targe
     });
 }
 
-async function getTargetIdFromFile(filePath) {
+async function getTargetIdFromConfigFile(filePath) {
     if (!await fileExists(filePath)) {
         return null;
     }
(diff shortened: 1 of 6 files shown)


libraries/utilities.js
```diff
@@ -254,6 +232,76 @@ async function getAnchorIdFromObjectFile(folderName) {
...
+                let targetUniqueId = await getTargetIdFromConfigFile(configFilePath);
+                console.log('TODO: cleanup config.info file instead of leaving it in the folder');
+                resolve(targetUniqueId);
+            } else {
+                reject('config.info not found at ' + configFilePath);
+            }
+        });
+
+        unzipperDat.on('progress', function (_fileIndex, _fileCount) {
+            // console.log('Extracted 3dt file ' + (fileIndex + 1) + ' of ' + fileCount);
```
ptc-mklaudiny, line 258: mentiones 3dt file not dat file.  Also, the comment few lines down is for 3dt

models/ObjectModel.js
```diff
@@ -15,6 +15,8 @@ function ObjectModel(ip, version, protocol, objectId) {
     this.objectId = objectId;
     // The name for the object used for interfaces.
     this.name = '';
+    // The UUID used internally by Vuforia for tracking
+    this.targetId = null;
```
ptc-mklaudiny, line 19: This is more conceptual question why Object does not have some equivalent of publicData/privateData. Essentially, a place were data specific to object type can be stored (eg. targetId). Just some types of objects have Vuforia targets associated with them.

server.js
```diff
@@ -753,6 +753,9 @@ async function loadObjects() {
                 const obj = objects[tempFolderName];
                 obj.ip = services.ip; // ip.address();
 
+                // update the targetId if needed
+                obj.targetId = await utilities.getTargetIdFromTargetDat(path.join(objectsPath, objectFolderList[i], identityFolderName, 'target'));
```
ptc-mklaudiny, line 757: Should there be a check whether fetching from dat file is necessary? Targetid might be in object.json already.

ptc-mklaudiny (approved):
Some questions but looks fine to me.

server.js
benptc: I think it's safer to re-read it from the dat file when the server starts, in case the user tampered with the filesystem and swapped out the target file while the server was offline.

models/ObjectModel.js
benptc: Yeah, it's probably worth refactoring some of the properties of the Object data model at some point. I'm hesitating to add publicData/privateData to objects yet, as there are a set of APIs available in toolbox that let you interact with those properties (for tools and nodes), and adding the corresponding APIs for objects is beyond the scope of what I want to do in this PR. If I leave out those APIs but put targetId in publicData anyways, it feels to me that the publicData of objects will be quite different in purpose/functionality compared to that of tools/nodes, which might be confusing. I'm open to more discussion but we might need another PR for that.

ptc-mklaudiny: I agree that this beyond the scope here, but I wanted to raise the question.

libraries/utilities.js
```diff
@@ -254,6 +232,74 @@ async function getAnchorIdFromObjectFile(folderName) {
...
+        unzipperDat.extract({
+            path: targetFolderPath,
+            filter: function (file) {
+                return file.type !== 'SymbolicLink' && file.filename.endsWith('info');
+            }
+        });
+    });
+};
+
+async function getTargetIdFromConfigFile(filePath) {
```
hobinjk-ptc, line 270: This function isn't guaranteed to return the correct resultId since xml2js Parser's parseString could call its callback asynchronously (even though it's sync right now, the way it's structured with a callback function could become async at any time).

I'd recommend making another function getTargetIdFromXML() that wraps only the parseString in a classic `new Promise(function (resolve, reject) {})`

libraries/utilities.js
```diff
@@ -280,24 +290,34 @@ async function getTargetIdFromConfigFile(filePath) {
         return null;
     }
 
-    let resultId = null;
+    return await queryXMLContents(contents, (xml) => {
```
hobinjk-ptc, line 293: If queryXMLContents throws due to the parse failing, this function will bubble the error up. I think the intended behavior is to wrap this `return` in a `try {} catch {}` that returns null if something explodes

benptc: Thank you for `catch`ing my last `try`, so to speak. Added that.

hobinjk-ptc: Ah yes, `finally`

hobinjk-ptc (approved):

PR #971, opened 2024-01-12
configure winston to ignore spammy decompress-zip warnings

benptc:
Not sure if there's a better approach for this, but currently I'm getting hundreds of warning lines from the decompress-zip library every time I use it to decompress the target.dat (which I do on server restart to ensure the file didn't change while the server was off), like:

```
2024-01-12T20:03:10.652Z - warn: Possibly unsupported ZIP platform type, 19 
2024-01-12T20:03:10.652Z - warn: Possibly unsupported ZIP platform type, 19 
2024-01-12T20:03:10.652Z - warn: Possibly unsupported ZIP platform type, 19 
...
2024-01-12T20:03:10.652Z - warn: Possibly unsupported ZIP platform type, 19
```
This seems to work for suppressing those messages.

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -40,7 +40,15 @@ const colorizedFormat = format.combine(
 
 const logger = createLogger({
     level: process.env.LOG_LEVEL || 'debug',
-    format: isMobile ? monochromeFormat : colorizedFormat,
+    format: winston.format.combine(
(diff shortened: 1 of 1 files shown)


hobinjk-ptc (approved):
Nice to have! And yeah, looks like this is the best way to do it

PR #973, opened 2024-01-24
APIs for ThingWorx tool / future data stream tools

benptc:
Adds some plumbing to the server to allow tools to call a REST APIs on the server to interact with hardware interfaces that provide configurable "data sources" and "data streams". For example, adding a new ThingWorx endpoint in the IoT tool would send a message to any hardware interfaces subscribed to these events, telling it to set up that data source (if it supports that type of data source). The hardware interface can then query the endpoint at a certain interval to get a list of "data streams" (endpoints for specific properties / sensor values) associated with that data source. The ThingWorx tool can query the server to get a list of these data streams, and upon choosing one, can post to the `/logic/bindNodeToDataStream` endpoint to tell the server to begin streaming values from a particular data stream to a particular node of a tool in perpetuity.


libraries/DataStreamInterface.js
```diff
@@ -0,0 +1,193 @@
+const { DataStreamServerAPI, NodeBinding } = require('@libraries/dataStreamInterfaces');
```
hobinjk-ptc, line 1: This could be simplified to ./dataStreamInterfaces

hobinjk-ptc (approved):
Nice comprehensive documentation!

PR #976, opened 2024-01-26
Mobile wrapping and no exit

hobinjk-ptc:
Following this, the iOS app can invoke index-nofork for a guaranteed* error free experience

*not enforceable in a court of law

### DIFF
--- a/index-nofork.js
+++ b/index-nofork.js
@@ -4,8 +4,13 @@ process.on('uncaughtException', (e) => {
     console.error('Uncaught server error', e);
 });
 
-try {
-    const _ = require('./server.js');
-} catch (e) {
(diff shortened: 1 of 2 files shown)


ptc-mklaudiny (changes requested):
I tested by calling new script in WebServerManager
        let srcPath = Bundle.main.path(forResource: "vuforia-spatial-edge-server/index-nofork.js", ofType: "")
        let nodeArguments:[String?] = ["node", "--jitless", srcPath]
   
But it does not lead to graceful shutdown or restart.

hobinjk-ptc:
Unfortunate! Some more in-depth approach will need to be taken then

PR #977, opened 2024-01-29
Feature: Synchronize filesystem with cloud proxy

hobinjk-ptc:
Pretty large feature at the core of the "run clones when the real server is offline" strategy. Corresponding API implementation is in the vuforia-spatial-cloud-proxy repo. Main idea is to run both the fs function and the cloud proxy's fetch-based implementation to keep the local and cloud mostly in sync. If local and cloud differs, then the server runs a big "checksum everything and upload/download what has differences"


dangond-ptc (approved):
I don't see anything obviously wrong here, though it feels very weird to me for this to be part of the edge-server instead of the edge-agent plugin (or rather, it feels strange that the edge-agent plugin isn't a native feature of the edge-server). Also feels wrong to store files directly in a DB, but from what I could tell, that's actually fine with Mongo?

The one real concern I have is the `write_file` and `read_file` endpoint, which, if not carefully protected, can allow for some nasty security vulnerabilities.

This code LGTM though, so approving.

hobinjk-ptc:
Oh, for clarification I've removed all the MongoDB code. It was an approach I used in the earlier stages of this project but didn't really scale since each edge server would need credentials for our db. The API on the cloud-proxy is currently backed by the filesystem but could use the same mongodb backend that used to be in this code.

`read_file` and `write_file` both have some amount of sanitation applied in the cloud-proxy.

Ideally edge-agent would be part of the server for sure, the tendrils of its responsibilities snake into every nook and cranny

PR #980, opened 2024-01-31
Offline clone improvements

hobinjk-ptc:


persistence/CloudProxyWrapper.js
```diff
@@ -9,13 +12,31 @@ class CloudProxyWrapper {
     connect() {
     }
 
+    localToRemote(path) {
+        if (path.startsWith(objectsPath)) {
+            let relPath = pathOps.relative(objectsPath, path);
+            if (!relPath.startsWith('..')) {
```
dangond-ptc, line 18: I believe `..` can be within a path, e.g. `folder/other/../file.ext`

persistence/CloudProxyWrapper.js
```diff
@@ -9,13 +12,31 @@ class CloudProxyWrapper {
...
+                return relPath;
+            }
+        }
+        return path;
+    }
+
+    remoteToLocal(path) {
+        if (path.startsWith('/') || path.startsWith('..')) {
```
dangond-ptc, line 26: Might want to also cover `~` at the start of the path

persistence/CloudProxyWrapper.js
```diff
@@ -9,13 +12,31 @@ class CloudProxyWrapper {
...
+    }
+
     apiBase() {
         const {getLoadedHardwareInterface} = require('../libraries/utilities.js');
         const edgeAgent = getLoadedHardwareInterface('edgeAgent');
         if (!edgeAgent || !edgeAgent.networkUUID || !edgeAgent.networkSecret) {
             throw new Error('bad edge-agent settings: ' + JSON.stringify(edgeAgent));
         }
-        const serverUrl = 'alpha.toolboxedge.net' || edgeAgent.serverUrl;
+        const serverUrl = edgeAgent.serverUrl || 'alpha.toolboxedge.net';
```
dangond-ptc, line 39: oops

persistence/CloudProxyWrapper.js
hobinjk-ptc: In theory `pathOps.relative` should normalize `path` to remove that sort of thing but that isn't explicit in its documentation so I'll get more paranoid with it

dangond-ptc: Oh I see, I was a bit confused by checking for `startsWith('..')` if relative was meant to remove that, but I assume it can start with `..` if it's in a higher relative directory. Looks good to me as is now that I understand how that interaction works.

persistence/CloudProxyWrapper.js
dangond-ptc: Is anything going on here with pathOps.relative that I'm not seeing?

persistence/CloudProxyWrapper.js
hobinjk-ptc: Adjusted this function and remoteToLocal to be a bit stronger and clearer in their guarantees

persistence/CloudProxyWrapper.js
hobinjk-ptc: Sort of, but I've adjusted the function to be clearer in what it's going for

dangond-ptc (approved):

PR #991, opened 2024-02-21
Download when step is complete

hobinjk-ptc:
This commit fell through the cracks. Note that this is also the pathway to downloading the partial checkpoint target.splat and later downloading the final (complete) file


benptc (approved):
Good catch!

PR #997, opened 2024-02-26
Enable https

ptc-rdeleeuw:
enables https on desktop targets while using http on mobile platforms

depends on:
https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/pull/633
https://github.com/ptcrealitylab/vuforia-spatial-remote-operator-addon/pull/156
https://github.com/ptcrealitylab/pop-up-onboarding-addon/pull/113
https://github.com/ptcrealitylab/vuforia-spatial-toolbox-ios-swift/pull/593


libraries/webInterface/gui/index.js
```diff
@@ -619,14 +619,14 @@ realityServer.updateManageObjects = function (thisItem2) {
...
-                    // let targetUrl = 'http://localhost:8080/obj/' + thisObject.name + '/target/target.jpg';
+                    // let targetUrl = (realityEditor.network.useHTTPS ? 'https' : 'http') + '://localhost:8080/obj/' + thisObject.name + '/target/target.jpg';
                     // thisObject.dom.querySelector(".target").style.backgroundImage = 'url("' + targetUrl + '")';
                     // thisObject.dom.querySelector(".target").style.backgroundSize = 'cover';
 
                     let ipAddress = realityServer.states.ipAdress.interfaces[realityServer.states.ipAdress.activeInterface];
                     // add Image for Target
                     if (thisObject.targetsExist.jpgExists) {
-                        thisObject.dom.querySelector('.objectTargetIcon').src = 'http://' + ipAddress + ':' + realityServer.states.serverPort + '/obj/' + thisObject.name + '/target/target.jpg';
+                        thisObject.dom.querySelector('.objectTargetIcon').src = (realityEditor.network.useHTTPS ? 'https' : 'http') + '://' + ipAddress + ':' + realityServer.states.serverPort + '/obj/' + thisObject.name + '/target/target.jpg';
```
hobinjk-ptc, line 629: Why use `useHTTPS` here and `location.protocol` above?

scripts/ci.sh
```diff
@@ -1,17 +1,19 @@
 #!/bin/bash -x
 
+./scripts/ci_generate_certificate.sh
+
 mkdir screenshots
 
 npm ci
 
-git clone https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface
+git clone --branch enable_https https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface
```
hobinjk-ptc, line 9: Note to revert this before merging

hobinjk-ptc (changes requested):
Overall LGTM! Great, comprehensive work!

Just have some minor changes I've PR-ed into this branch with some improvements I made locally to the cert gen process

server.js
```diff
@@ -100,9 +100,12 @@ const globalVariables = {
         udp: false,
         web: false,
         system: false
-    }
+    },
+    useHTTPS: !(isLightweightMobile || isStandaloneMobile) // on mobile devices node.js doesn't fully support HTTPS
```
hobinjk-ptc, line 104: Thinking through the workflow if we ever want to get a customer to run the server for a local access usecase (something that one of them just requested) it would be better for the server to gracefully fall back to http if the certs aren't available instead of forcing https on for all desktop users. This would take the form of checking whether cert.pem/key.pem exists, warning if not on mobile but still setting useHTTPS to false.

ptc-rdeleeuw: My hope is that https will be come available on node.js for ios and unify the server to always run in a secure context (this solves problems with [features that want https](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts/features_restricted_to_secure_contexts)). if we enable this fallback that will make this unification much harder to do later. @Benolds What do you think?

hobinjk-ptc: Hmm, I guess I'm convinced. I think I'm okay with targeting HTTPS everywhere and the real target is documentation and messaging around certificate generation and trusting such that it's not a massive pain

PR #998, opened 2024-02-26
Recorder is fast now

hobinjk-ptc:
tl;dr (more in commit message): diffing is 16x faster

### DIFF
--- a/libraries/recorder.js
+++ b/libraries/recorder.js
@@ -230,12 +230,12 @@ recorder.applyDiff = function(objects, diff) {
 
 /**
  * @param {object} cursorObj - view into recorder.object
- * @param {object} cursorObj - view into recorder.objectOld
+ * @param {object} cursorObjOld - view into recorder.objectOld
  * @param {object} cursorTime - view into the time object
  *                 between recorder.object and recorder.objectOld
  * @return {boolean} whether the recursion detected a change
  */
-recorder.recurse = function (cursorObj, cursorOld, cursorTime) {
+recorder.recurse = function (cursorObj, cursorObjOld, cursorTime) {
     let altered = false;
     for (const key in cursorObj) { // works for objects and arrays
         if (!cursorObj.hasOwnProperty(key)) {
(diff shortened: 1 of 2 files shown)


libraries/recorder.js
```diff
@@ -168,20 +172,29 @@ recorder.update = function() {
...
+ * @param {number} targetTime
+ * @param {object|undefined} checkpoint
+ */
 recorder.replay = function(timeObject, targetTime, checkpoint) {
-    let objects = {};
-    let currentTime = 0;
-    if (checkpoint) {
+    let times = Object.keys(timeObject).map(t => parseInt(t));
+
+    let objects = JSON.parse(JSON.stringify(timeObject[times[0]]));
```
ptc-mklaudiny, line 184: Is stringify -> parse need for the undefined/null issue or is there other reason?

libraries/recorder.js
```diff
@@ -208,62 +229,63 @@ recorder.applyDiff = function(objects, diff) {
 };
 
 /**
- * @param {object} obj - view into recorder.object
- * @param {object} objectInTime - object to store any values that change
+ * @param {object} cursorObj - view into recorder.object
+ * @param {object} cursorObj - view into recorder.objectOld
```
ptc-mklaudiny, line 233: How about cursorObjOld ?

libraries/recorder.js
```diff
@@ -208,62 +229,63 @@ recorder.applyDiff = function(objects, diff) {
...
+            if (item !== cursorOld[key]) {
+                cursorTime[key] = item;
+                altered = true;
             }
 
-            oldItem[key] = thisValue;
+            cursorOld[key] = item;
         }
     }
+    return altered;
```
ptc-mklaudiny, line 288: recorder.getItemFromArray below can be removed

tests/recorder.test.js
```diff
@@ -0,0 +1,107 @@
+/* global test, expect */
+
+const recorder = require('../libraries/recorder.js');
```
ptc-mklaudiny, line 3: So nice to see a unit test! :)

tests/recorder.test.js
```diff
@@ -0,0 +1,107 @@
...
+    objects.a[1] = 12;
+    recorder.saveState(1);
+
+    expectedTimeObject[1] = { a: [ undefined, 12 ] };
+    expect(recorder.timeObject).toEqual(expectedTimeObject);
+
+    // Delete b
+    delete objects.b;
+    recorder.saveState(2);
+    // No modification, deletions aren't reflected in diff
```
ptc-mklaudiny, line 61: Now I realise that we do not get the exact state of var objects at a given time by replay mechanism. In practice, how do we then distinguish whether an object did not change for some time or whether it is already deleted?

ptc-mklaudiny (commented):
I never quite understood the keyString 'path' addressing of items. This is clearer to me now.
16x speedup is great!

libraries/recorder.js
hobinjk-ptc: Thanks!

tests/recorder.test.js
hobinjk-ptc: I believe most cases use an idle timeout type approach "if object not updated in X milliseconds then it's deleted" but this is definitely an area where I don't remember the exact particulars

libraries/recorder.js
ptc-mklaudiny: ah, right.

ptc-mklaudiny (approved):

PR #1000, opened 2024-02-28
Move cert gen scripts into scripts dir; Make bash cert gen script fancier; Update README.md

hobinjk-ptc:
Notably the updates to the bash cert gen script allow you to add your local ip address to the cert.conf file, add the generated certs to Keychain, then never have to click to manually trust the certs in chrome/safari


scripts/generate_cert.ps1
ptc-rdeleeuw, line 1: I will sync this script after merging with the shell variant

PR #1003, opened 2024-02-29
New tools now initialize properly on slow networks

dangond-ptc:
New tools now initialize properly on slow networks, previous method sent subscribe method before socket was ready, resulting in no subscription being created.

This fixes the following bugs:
• On slow networks, placing a tool fails to load the tool, despite creating it on the server
• On slow networks, new tools added by other users fail to subscribe to publicData messages, and can only write, not read updates

To reproduce said network conditions, set network speed to Slow 3G in Chrome's network pane, Fast 3G reproduces these bugs only sometimes.

### DIFF
--- a/libraries/objectDefaultFiles/object.js
+++ b/libraries/objectDefaultFiles/object.js
@@ -909,20 +909,22 @@
          * Subscribes this socket to data values being written to nodes on this frame
          */
         this.sendRealityEditorSubscribe = function () {
-            var timeoutFunction = function() {
-                if (spatialObject.object) {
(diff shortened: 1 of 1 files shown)


dangond-ptc:
Main issues were that `spatialObject.object` existing was insufficient, resulting in the subscribe message never being sent, and that it gives up after a second.

hobinjk-ptc (approved):

dangond-ptc:
It seems very possible that this is also the source of the "iPad needs to draw first" bug we've been seeing on the demo machine

benptc (approved):
Great catch!

PR #1005, opened 2024-02-29
Recorder: Only save files with differences

### LINKED ISSUE #1004: Prevent creation of new log files when no changes have occured

hobinjk-ptc:
Reduces the recorder's impact on the filesystem by only creating a new
log file if that log file contains differences from the last (or if it's the first call to persist after starting the server).

Also minor fix for potential bug (that doesn't happen in practice)

Close #1004

### DIFF
--- a/libraries/recorder.js
+++ b/libraries/recorder.js
@@ -106,25 +106,57 @@ recorder.getAndGuaranteeOutputFilename = function(logName) {
 };
 
 /**
- * Persist current state to file, skipping if no differences recorded since last persist
- * @return {Promise<string|null>} log file name
+ * Get all log data that needs to be persisted and how to identify it. Mark all
+ * of this claimed log data as persisted, removing it from the timeObject
+ * @return {{logName: string, timeObjectStr: string}|null}
  */
-recorder.persistToFile = function () {
+recorder.getAndMarkPersistedCurrentLog = function() {
     // recorder.objectOld is the latest state persisted in timeObject
     // Because we're using objectOld as the basis for each log file, having
     // only one entry in timeObject means `object` never had a difference from
     // `objectOld` and we can skip writing to the file.
-    if (Object.keys(recorder.timeObject).length <= 1 && hasPersistedToFile) {
-        return Promise.resolve(null);
+    let recordedTimes = Object.keys(recorder.timeObject);
+
+    // Nothing recorded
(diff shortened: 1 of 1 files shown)


libraries/recorder.js
```diff
@@ -83,11 +87,26 @@ recorder.getAndGuaranteeOutputFilename = function(logName) {
...
 recorder.persistToFile = function () {
+    // recorder.objectOld is the latest state persisted in timeObject
+    // Because we're using objectOld as the basis for each log file, having
+    // only one entry in timeObject means `object` never had a difference from
+    // `objectOld` and we can skip writing to the file.
+    if (Object.keys(recorder.timeObject).length <= 1 && hasPersistedToFile) {
+        return Promise.resolve(null);
+    }
+
+    hasPersistedToFile = true;
```
ptc-mklaudiny, line 103: This is set just once for the first time after start. Is the intention to skip creation of new log file when there was no change at all since the last call of persistToFile()? Essentially, delaying the creation of next log file to some future call of persistToFile when there are some new changes to store.

libraries/recorder.js
```diff
@@ -117,16 +136,27 @@ recorder.persistToFile = function () {
...
+
     let logName = recorder.getCurrentLogName();
     let timeObjectStr = JSON.stringify(recorder.timeObject);
-    recorder.objectOld = {};
-    recorder.timeObject = {};
+    recorder.timeObject = {
+        [Date.now()]: JSON.parse(JSON.stringify(recorder.objectOld)),
```
ptc-mklaudiny, line 152: What if someone starts to use saveState(time) with some time param 'not aligned' with current time?
Maybe Date.now() could be replaced by the last time stamp in the timeObject which is going to be saved.

ptc-mklaudiny (commented):
Overall, it makes sense to me. 
Could you add documetation to the rest of functions ? In the interest of  continuous improvement of understandability of our code :)

libraries/recorder.js
hobinjk-ptc: Yes, that's the goal with this overall pr

libraries/recorder.js
hobinjk-ptc: That feels suitably elegant, modified accordingly

ptc-mklaudiny (approved):
Thanks for docs. LGTM

PR #1009, opened 2024-03-04
Add documentation about trusting self-signed certs on iOS

hobinjk-ptc:


ptc-rdeleeuw:
this only concerns the documentation but there is no mention of how to start the python webserver mentioned in the documentation. I know this is easy to do, but we might want to add a shell script to the scripts folder or add the commandline starting the python server in the documentation it self

docs/ios-cert-install.md
ptc-rdeleeuw, line 1: this only concerns the documentation but there is no mention of how to start the python webserver mentioned in the documentation. I know this is easy to do, but we might want to add a shell script to the scripts folder or add the commandline starting the python server in the documentation it self

hobinjk-ptc: Oh sorry, you can see https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/1009/files#diff-95c07260d1a629a46230701e387127ad04f277913dc2b8cb78e955a460edd17cR14 is where I meant to put the snippet

hobinjk-ptc: (added)

ptc-rdeleeuw (approved):

PR #1011, opened 2024-03-06
Add --insecure command line option

hobinjk-ptc:


ptc-rdeleeuw (changes requested):
Please adjust jest tests suite for github actions to run the same tests for the insecure version.

ptc-rdeleeuw (approved):
Thanks good work. Are there more repositories with tests that we will have to duplicate?

hobinjk-ptc:
Other repositories' tests are all self-contained (and usually just linting)

hobinjk-ptc:
Decided to invert the logic since that way the desktop and iOS behavior will match and the https will be more of a special case for webxr/locally-deployed Teams

ptc-rdeleeuw (approved):
Since it has been decided to add this switch I don't care which version is the default http or https. That is provided the tests written for the code tests both versions which seems to be the case. So my initial approval doesn't change. Again good work, thumbs up.

hobinjk-ptc:
Thanks!

PR #1017, opened 2024-03-15
Empty worlds, checkTargetFiles, and renderMode

benptc:
- Adds POST /object/:id/renderMode - to toggle a world's render mode from mesh to ai
- Adds POST /object/:id/checkFileExists/:filePath - to check existence of a single file within the object's .identity directory (e.g. targets/target.glb) (instead of continuously checking the file itself to see when the mesh is available, which results in an extra download of a potentially large file)
- Adds POST /object/:id/checkTargetFiles/ - to get the existence of each of the target files (xml, dat, splat, 3dt, glb, jpg) in a single fetch
- Implements new uploadTarget function that refactors/imitates a lot of logic from the /content/:id upload route, that isn't actually used because it's not supported by toolsocket, but I left it there because the content route pains me and we might want something like this in the future
- World objects are initialized by default, don't auto-generate an XML file on creation, and can be clicked on in the 8080 interface to go the metaverse page even without target data attached
- Better cleanup the target folder after uploading a target zip file


benptc:
Oops I'll fix the errors, just trying to get some of my code merged into main instead of relying on alpha forever...

dangond-ptc (approved):

hobinjk-ptc (approved):

hobinjk-ptc (changes requested):
The test failure is the rather weird:
```
 FAIL tests/data-streams.test.js
  ● Test suite failed to run

    TypeError: hardwareAPI.setup is not a function

      662 | };
      663 | // set all the initial states for the Hardware Interfaces in order to run with the Server.
    > 664 | hardwareAPI.setup(objects, objectLookup, knownObjects, socketArray, globalVariables, __dirname, objectsPath, nodeTypeModules, blockModules, services, serverPort, hardwareAPICallbacks, sceneGraph, worldGraph);
          |             ^
      665 |
      666 | const utilitiesCallbacks = {
      667 |     triggerUDPCallbacks: hardwareAPI.triggerUDPCallbacks

      at Object.setup (server.js:664:13)
      at Object.require (controllers/object.js:7:16)
      at Object.require (libraries/hardwareInterfaces.js:26:26)
      at Object.require (libraries/DataStreamInterface.js:2:16)
      at Object.require (tests/data-streams.test.js:10:29)
      ```

File: controllers/object.js [hunk 1]
```diff
@@ -4,6 +4,8 @@ const formidable = require('formidable');
 const utilities = require('../libraries/utilities');
 const {fileExists, unlinkIfExists, mkdirIfNotExists} = utilities;
 const {startSplatTask} = require('./object/SplatTask.js');
+const server = require('../server');
```
hobinjk-ptc, line 7: I don't think this kind of circular import is well-liked by the server but I might be wrong here

hobinjk-ptc (approved):
LGTM!

PR #1031, opened 2024-04-02
Toolsocket updates

dangond-ptc:
- Adapted codebase to use ToolSocket v2.0.0's API.


dangond-ptc:
I'm still working on tracking down the cause of the failing tests.
This PR should not be merged until the ToolSocket repo officially updates to 2.0.0.

dangond-ptc:
Ok, after much investigation: the current code passes all tests on a clean install with the right toolsocket dependency. The issues I was having are:
1. chromium-browser must be installed for puppeteer to work, can be installed on MacOS with `brew install chromium` and Ubuntu with `sudo apt-get install chromium-browser`
2. The test `tests/data-streams.test.js` fails on my laptop for some unknown reason. This is true for the current `main` branch as well. I tested on an Ubuntu machine instead, and all tests pass there. Nothing else seems to be different between the two machines other than my laptop has a `Documents/spatialToolbox` directory, which, from what I can tell, should not affect this test.

libraries/utilities.js
```diff
@@ -75,7 +75,7 @@ const {identityFolderName} = require('../constants.js');
 var hardwareIdentity = path.join(objectsPath, identityFolderName);
 
 let socketReferences = {
-    realityEditorUpdateSocketArray: null
+    realityEditorUpdateSocketSubscriptions: null
 };
 
 let ioReference = null;
```
hobinjk-ptc, line 81: Can drop this variable and its use in utilities.setup()

server.js
```diff
@@ -4170,18 +4203,18 @@ async function deleteObject(objectKey) {
     sceneGraph.removeElementAndChildren(objectKey);
 }
 
-function sendMessagetoEditors(msgContent, sourceSocketID) {
-    for (var thisEditor in realityEditorSocketArray) {
-        realityEditorSocketArray[thisEditor].forEach((thisObj) => {
-            if (typeof sourceSocketID !== 'undefined' && thisEditor === sourceSocketID && msgContent.object === thisObj.object && msgContent.frame === thisObj.frame) {
+function sendMessageToEditors(msgContent, sourceSocket) {
```
hobinjk-ptc, line 4206: Great improvement in readability!

hobinjk-ptc (approved):

hobinjk-ptc:
(fix lint errors before merging)

server.js
```diff
@@ -3474,17 +3476,23 @@ function socketServer() {
             }
 
             if (doesObjectExist(msgContent.object)) {
-                if (!realityEditorSocketArray[socket.id]) realityEditorSocketArray[socket.id] = [];
+                if (!realityEditorSocketSubscriptions.some(s => s.socket === socket)) {
+                    const subscriptions = [];
+                    subscriptions.socket = socket;
+                    realityEditorSocketSubscriptions.push(subscriptions);
```
benptc, line 3482: Little bit confused about adding a socket property to an array, rather than using an object, is there a reason to structure it like this? (I'm not always up-to-date with clean coding conventions so I figure you have a valid reason, just curious)

README.md
```diff
@@ -76,6 +76,12 @@ consistently styled and functional. If you want to see the results of this
 testing locally, you can follow
 [this Github Actions workflow](https://github.com/ptcrealitylab/vuforia-spatial-edge-server/blob/master/.github/workflows/nodejs.yml).
 
+```bash
+git clone --recurse-submodules https://github.com/ptcrealitylab/vuforia-spatial-edge-server.git
+cd vuforia-spatial-edge-server
+./scripts/ci.sh
+```
```
benptc, line 83: Thanks for updating the README, too!

benptc (approved):
LGTM

server.js
dangond-ptc: Looking at it now, I cannot come up with a good answer for this. I'll push a commit with a more sensible arrangement!

dangond-ptc:
@benptc Added a new commit regarding the socket property on the arrays. Lmk if that's better and looks good!

PR #1040, opened 2024-04-11
AI Chatbot

Steve-KX-RL:
1. Have a (somewhat) useable conversation with your Azure AI chatbot, after inputting the credentials.
2. Get a list of historical & new events from user-interface, to feed into AI chatbot.
3. Decide the category of user question (summary/debug/tool/pdf/tool content/not relevant), and give different answers based on that.
4. Also need user-interface and remote-operator to work.


server.js
```diff
@@ -1952,6 +1952,171 @@ function objectWebServer() {
         res.sendStatus(200); // OK
     });
 
+    function parseCategory(result) {
```
hobinjk-ptc, line 1955: This addition would be better organized if it were in its own file. One example of this is objectRouter and other routers defined above

hobinjk-ptc (changes requested):

package-lock.json
Steve-KX-RL, line 1: Just realized I accidentally committed this file. Should I make another commit to revert it back to what it was before?

hobinjk-ptc: No, it looks like this properly reflects the addition of `@azure/openai` to the dependencies

hobinjk-ptc (approved):

benptc (approved):
Great start – LGTM for now. Looks like this is a construction zone and there's plenty of work that can still be done, but it's cleanly contained in routers/ai.js so that should be fine for now.

PR #1042, opened 2024-04-15
Add increased max sizes to target/content forms

hobinjk-ptc:


dangond-ptc (approved):
LGTM, but should we consider increasing the size beyond 300MB if we frequently seem to be making scans in the 200MB to 300MB range? Would be nice to have more leeway, especially if the scan resolution ever changes in the future.

hobinjk-ptc:
Fair! I was beginning to stew on 300 MB being the new 10 MB. May as well go around and make sure we'll never have this problem again

PR #1052, opened 2024-04-24
adds authentication and error handling to SplatTasks

benptc:
@hobinjk-ptc this may be a bit of a haphazard update to your SplatTask module, feel free to propose a cleaner implementation. Passes JWT from userinterface to the splat training server and passes error messages back to userinterface if not properly authenticated.

### DIFF
--- a/controllers/object/SplatTask.js
+++ b/controllers/object/SplatTask.js
@@ -18,10 +18,13 @@ const SPLAT_HOST = 'change me:3000';
 class SplatTask {
     /**
      * @param {ObjectModel} object
+     * @param {string} userJWT
      */
-    constructor(object) {
+    constructor(object, userJWT) {
         this.object = object;
+        this.userJWT = userJWT;
         this.gaussianSplatRequestId = null;
+        this.error = null;
         if (this.object.gaussianSplatRequestId) {
             this.gaussianSplatRequestId = this.object.gaussianSplatRequestId;
         }
@@ -62,13 +65,23 @@ class SplatTask {
                 method: 'POST',
                 headers: {
                     ...form.getHeaders(),
+                    Authorization: `Bearer ${this.userJWT}`,
                 },
                 body: form,
             });
 
-            const gaussianSplatRequestId = await res.text();
-            this.object.gaussianSplatRequestId = gaussianSplatRequestId;
-            this.gaussianSplatRequestId = gaussianSplatRequestId;
+            let responseText = await res.text();
+            try {
+                // response is a string if successful, a JSON {error: 'reason'} if not
+                let errorJSON = JSON.parse(responseText);
+                this.error = errorJSON.error;
+                return null;
+            } catch (e) {
+                const gaussianSplatRequestId = responseText;
+                this.object.gaussianSplatRequestId = gaussianSplatRequestId;
+                this.gaussianSplatRequestId = gaussianSplatRequestId;
+            }
+
         }
 
         this.openSocket();
@@ -102,12 +115,13 @@ class SplatTask {
     }
 
     /**
-     * @return {{done: boolean, gaussianSplatRequestId: string|undefined}} splat status
+     * @return {{done: boolean, gaussianSplatRequestId: string|null, error: string|null}} splat status
      */
     getStatus() {
         return {
             done: this.done,
             gaussianSplatRequestId: this.gaussianSplatRequestId,
+            error: this.error,
         };
     }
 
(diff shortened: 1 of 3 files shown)


controllers/object/SplatTask.js
```diff
@@ -62,13 +65,23 @@ class SplatTask {
...
+                    Authorization: `Bearer ${this.userJWT}`,
                 },
                 body: form,
             });
 
-            const gaussianSplatRequestId = await res.text();
-            this.object.gaussianSplatRequestId = gaussianSplatRequestId;
-            this.gaussianSplatRequestId = gaussianSplatRequestId;
+            let responseText = await res.text();
+            try {
```
dangond-ptc, line 74: I think this would be cleaner via something like `if (res.status === 200) doNormalStuff else doErrorStuff`

routers/object.js
```diff
@@ -452,7 +452,13 @@ router.post('/:objectName/requestGaussianSplatting/', async function (req, res)
     // splat status (commonly referred to as "splattus") is
     // {done: boolean, gaussianSplatRequestId: string|undefined}
     try {
-        const splatStatus = await objectController.requestGaussianSplatting(req.params.objectName);
+        let userJWT = null;
+        const credentials = req.headers.authorization;
+        if (credentials && credentials.includes(' ')) {
+            // The credentials will be in the format "Bearer <JWT>", can be split to extract the JWT
+            userJWT = credentials.split(' ')[1];
```
dangond-ptc, line 459: Instead of parsing the JWT out and then re-combining it with 'Bearer: ', we can just pass `credentials` through the pipeline all the way to the fetch request, then simply `Authorization: credentials` there.

controllers/object/SplatTask.js
```diff
@@ -62,13 +65,23 @@ class SplatTask {
...
                     ...form.getHeaders(),
+                    Authorization: `Bearer ${this.userJWT}`,
                 },
                 body: form,
             });
 
-            const gaussianSplatRequestId = await res.text();
-            this.object.gaussianSplatRequestId = gaussianSplatRequestId;
-            this.gaussianSplatRequestId = gaussianSplatRequestId;
+            let responseText = await res.text();
```
hobinjk-ptc, line 73: This is another place where an error could happen and should probably be wrapped in a try/catch of its own

controllers/object/SplatTask.js
```diff
@@ -62,13 +65,23 @@ class SplatTask {
...
+            try {
+                // response is a string if successful, a JSON {error: 'reason'} if not
```
hobinjk-ptc, line 75: The API being shaped like this is less than ideal. I'm guessing it's out of scope to change it now but expecting a parse exception isn't good

hobinjk-ptc (approved):

controllers/object/SplatTask.js
```diff
@@ -11,18 +11,22 @@ const {fileExists} = utilities;
 
 const SPLAT_HOST = 'change me:3000';
 
+if (SPLAT_HOST.includes('change me')) {
+    console.warn('Edit the SPLAT_HOST if you want to enable Gaussian Splat training');
```
hobinjk-ptc, line 15: Nice change!

controllers/object/SplatTask.js
```diff
@@ -143,12 +158,7 @@ class SplatTask {
     async download() {
         const splatPath = path.join(objectsPath, this.object.name, identityFolderName, 'target', 'target.splat');
         try {
-            let res = await fetch(`http://${SPLAT_HOST}/downloads/${this.gaussianSplatRequestId}`, {
-                method: 'GET',
-                headers: {
-                    Authorization: `Bearer ${this.userJWT}`
-                }
-            });
+            let res = await fetch(`http://${SPLAT_HOST}/downloads/${this.gaussianSplatRequestId}`);
```
hobinjk-ptc, line 161: Does the GET not require authorization? That feels like another area for future improvement (as well as the more fundamental switching SPLAT_HOST to https)

hobinjk-ptc (approved):

dangond-ptc (approved):

controllers/object/SplatTask.js
dangond-ptc: I've only added authorization to the POST route since that's the expensive route, but I can add the auth-checking middleware to the other routes as well.

benptc: I checked with Dan – doesn't require authorization. I had added it as I thought it fixed a previous problem, then realized that my problem was just turning zscaler on/off and that this authorization wasn't needed

hobinjk-ptc: Yeah, definitely a nice-to-have in the future but not critical in the short term

PR #1074, opened 2024-06-06
AI function calling, state summary, and spatial references

benptc:
Adds `LanguageInterface.js` and `SpatialApplicationAPI.js` to objectDefaultFiles, which tools can use to declare their capabilities and contents to make them available to the AI interface.

Updates the routers/ai.js with a new /v2/query route, a /continue-query route, and a /process-state-with-prompts route, which together can be used to trigger complex sequences of function calling and question-and-answer responses from the GPT system in response to the state of the toolbox session and its tools.

Part of the https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/pull/725 set of PRs.


routers/ai.js
```diff
@@ -32,138 +15,272 @@ router.post('/init', async function(req, res) {
...
+    functionGuidelines: `The user may make a request that requires one or more function calls to fulfill it. Function calls are expensive. Before performing a function call, please prepare a plan of the sequence of function calls that you will need to perform to fulfill the request, and ask the user to verify the plan and give you explicit permission to call them. Upon receiving permission, please perform the sequence of function calls sequentially until all of them have been completed, then tell the user the results when all have been completed.`,
+    // functionGuidelines: `Function calls are triggered by explicit user requests or commands implying a series of actions, like "draw a blue square of size 1000". Interpret such commands as complete tasks to be executed in sequence, seeking clarification only if the user's intent or actions are unclear.`,
```
benptc, line 88: Note to anyone testing the system (@Steve-KX-RL , @hobinjk-ptc , @dangond-ptc) - you may want to swap the function guidelines back to the commented version, or to something like:

functionGuidelines: `The user may make a request that requires one or more function calls to fulfill it. Function calls are expensive. First decide if only one function call would be needed to fulfill the request. If so, make the request without asking for permission. But if the request requires multiple function calls, before performing the function calls, please prepare a plan of the sequence of function calls that you will need to perform to fulfill the request, and ask the user to verify the plan and give you explicit permission to call them. Upon receiving permission, please perform the sequence of function calls sequentially until all of them have been completed, then tell the user the results when all have been completed.`

...as the current version requires a confirmation step before every API call, which helps to structure its thoughts about complex sequences of calls, but gets in the way of simpler tasks.

benptc:
Closing this, as changes have been refactored to the spatial-ai-addon. New (simpler) PRs for server and userinterface have been opened: https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/1093 and https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/pull/737

PR #1079, opened 2024-06-11
Add analyticsSetSensor API

hobinjk-ptc:
@benptc This is what was missing from other branches


benptc (approved):
Thanks, this fixes it for me 👍

PR #1125, opened 2024-08-23
Optimize batched updates

benptc:
Creates a `BatchedUpdateAggregator` class to optimize the `/batchedUpdate` messages that the clients use to synchronize object/tool/node positions, e.g. when avatar objects move around or when you drag tools around.

This class pushes each message batch into a buffer that gets consolidated over an interval and sent out to all clients at the end of the interval. The interval is calculated dynamically based on the number of clients sending messages in the past few seconds. It starts low (30fps if there are less than 4 clients communicating) and decreases quadratically to 1fps by the time there are 15 or more clients communicating over the recent time window.

Empirically, if I open 15 remote operator windows on my (luckily powerful enough to handle it) MacBook, and say "All Follow Me" on one of them, and them press the O key to orbit, and then start dragging a tool icon to reposition it – the number of socket messages sent through the server used to be 600+ per second, now it is reduced to maybe 50 per second. In general, the traffic seems to be reduced by about a factor of N, where N is the number of clients. Though I cap it at 1 fps as the lowest, so I'm not sure what would happen if there are literally 50 or 100 clients connected, but that seems unlikely.

I've also made some optimizations to the clients so that they send fewer messages, which I'll add in a linked PR.


libraries/BatchedUpdateAggregator.js
```diff
@@ -0,0 +1,161 @@
...
+    getPeakClientCount() {
+        return Math.max(...this.activeClientsHistory, 1);
+    }
+
+    adjustAggregationInterval() {
+        const peakClientCount = this.getPeakClientCount();
+
+        // Set your desired BandwidthCap (number of messages per time window) and TimeWindow (e.g., 5 seconds)
+        const BandwidthCap = 1000;  // Example value, adjust based on your capacity
+        const TimeWindow = 5000; // 5 seconds in milliseconds
```
dangond-ptc, line 120: I would use ALL_UPPERCASE for these two, otherwise they look like classes imo

dangond-ptc (commented):
LGTM! Small nit included.

